PR #37, opened 2021-05-22
Add usage tracking through dedicated User-Agent header

nfx:
This PR will distinguish this scripts from other Python scripts, that run against Databricks REST API


mrchristine (approved):
lgtm, +1

PR #54, opened 2021-09-01
updated readme

ahoughro:
I attempted to add some initial setup steps, and modified some commands that were missing the --profile information and attempted to provide navigation related to errors that can come up from using the tool.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,6 +12,33 @@ Support for Windows is work in progress to update all paths to use pathlib resol
 This package uses credentials from the 
 [Databricks CLI](https://docs.databricks.com/user-guide/dev-tools/databricks-cli.html)  
 
+**Pre-Requisites**
+To use this migration tool, you'll need:
+*An environment running linux with python, pip, git, and the databricks CLI installed.
+*Admin access to both the old and new databricks accounts.
+
+Generate Access Tokens for both the old and new databricks accounts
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -107,6 +134,16 @@ This will export the following:
 ```bash
 python import_db.py --profile NEW_DEMO --clusters
 ```
+If you experience errors when you try to import the clusters, it may be that you need to modify the clusters file from the logs directory to include the new instance profile if it's not the same as the one in the old databricks account.
+
+To make changes to a cluster name to match the new databricks account you must edit the clusters log file after export.  You do this by looking at the clusters file and identifying the old cluster instance profile which will include the old account number and the name of the instance profile.
+OLD profile text from an AWS Databricks account:
+arn:aws:iam::111111111111:instance-profile/profileName
```
mrchristine, line 141: Great suggestion here. I do support an account id change w/ `[--update-account-id UPDATE_ACCOUNT_ID] [--old-account-id OLD_ACCOUNT_ID]` but this does not handle the `profileName` change that you mentioned. It sounds like a common use case for people to change role names. I'll add a new option to support that as well.

mrchristine (approved):
lgtm. Once of the suggestions is a feature request that all add to the tool.

PR #61, opened 2021-09-14
added --overwrite option for --import-home for a user

rohitkn:
Tested with basic inputs and works for import

### DIFF
--- a/dbclient/WorkspaceClient.py
+++ b/dbclient/WorkspaceClient.py
@@ -52,6 +52,7 @@ class WorkspaceClient(ScimClient):
         :return: return the full input args to upload to the destination system
         """
         is_source_format = self.is_source_file_format()
+        is_overwrite = self.is_overwrite()
         fp = open(full_local_path, "rb")
         (nb_path_dest, nb_type) = os.path.splitext(nb_full_path)
         in_args = {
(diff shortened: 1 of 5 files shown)


import_db.py
```diff
@@ -175,6 +175,7 @@ def main():
     if args.import_home:
         username = args.import_home
         print("Importing home directory: {0}".format(username))
+        client_config['overwrite'] = args.overwrite
```
mrchristine, line 178: We should not add the overwrite flag here, it should be added to `build_client_config()` in the parser.py file. If the arg exists, we add it to the client configs. This would allow us to re-use this flag for `--import-home` or general workspace imports.

dbclient/dbclient.py
```diff
@@ -33,6 +33,11 @@ def __init__(self, configs):
         self._is_verbose = configs['verbose']
         self._verify_ssl = configs['verify_ssl']
         self._file_format = configs['file_format']
+
+        self._overwrite = 'false'
+        if('overwrite' in configs):
+            self._overwrite = str(configs['overwrite']).casefold()
```
mrchristine, line 39: The config should store a boolean, so we should store a boolean here. There's no need to cast to string here.

dbclient/WorkspaceClient.py
```diff
@@ -52,6 +52,7 @@ def get_user_import_args(self, full_local_path, nb_full_path):
         :return: return the full input args to upload to the destination system
         """
         is_source_format = self.is_source_file_format()
+        is_overwrite = self.is_overwrite()
```
mrchristine, line 55: Let's collapse these 2 lines into a single line.

mrchristine:
I've added this capability across all workspace imports. We can close this PR.

PR #65, opened 2021-10-01
Pipeline interface draft

yubing:


pipeline/pipeline.py
```diff
@@ -0,0 +1,58 @@
...
+class Pipeline:
+    """Class that coordinates run of a group of tasks, which form a DAG based on dependencies
+    defined in add_task().
+
+    The pipeline has built-in checkpoint i.e it will skip all complete tasks upon restart."""
+
+    @dataclass
+    class Node:
+        task: AbstractTask = None
+        children = []
```
kevinkim9264, line 23: can you leave a comment for children? Is this the list of tasks that are executed after this Node(task) is done?

pipeline/pipeline.py
```diff
@@ -0,0 +1,58 @@
...
+
+    def __init__(self, checkpoint: Optional[str]):
+        self._source = self.Node()
+        self._checkpoint = checkpoint
+
+    def add_task(self, task: AbstractTask, parents: Optional[List[Node]] = None):
```
kevinkim9264, line 29: According to the examples down below, it seems like this method should return the `node`?

pipeline/pipeline.py
```diff
@@ -0,0 +1,58 @@
...
+    The pipeline has built-in checkpoint i.e it will skip all complete tasks upon restart."""
+
+    @dataclass
+    class Node:
+        task: AbstractTask = None
+        children = []
+
+    def __init__(self, checkpoint: Optional[str]):
+        self._source = self.Node()
+        self._checkpoint = checkpoint
```
kevinkim9264, line 27: so will the constructor and user of this pipeline instance use this checkpoint to add the tasks accordingly?

pipeline/pipeline.py
yubing: Yes, good catch!

pipeline/pipeline.py
yubing: Done. PTAL

pipeline/pipeline.py
yubing: The checkpoint here is actually a path for checkpointing. It's too early to discuss it right now. I changed it to a high level working_dir and added some comments.

migration_pipeline.py
```diff
@@ -0,0 +1,37 @@
...
+@dataclass
+class PipelineConfig:
+    """Configuration of the workspace migration pipeline."""
+    user: str
+    workspace: str
+    base_dir: str
+    session: str = ''
+
+
+def generate_session() -> str:
```
souravkhandelwal09, line 14: what does the session signify here ? Is it user session token used ?

maybe you can write a comment to describe it

kevinkim9264 (approved):

migration_pipeline.py
yubing: Done.

pipeline/pipeline.py
```diff
@@ -0,0 +1,53 @@
...
+        """
+        self._source = self.Node()
+        self._working_dir = working_dir
+        self._tasks = []
+
+    def add_task(self, task: AbstractTask, parents: Optional[List[Node]] = None) -> Node:
+        node = self.Node(task)
+        if not parents:
+            parents = [self._source]
+        for parent in parents:
```
kevinkim9264, line 39: Should we enforce that the passed in parents' tasks indeed exist in self._tasks?

pipeline/pipeline.py
```diff
@@ -0,0 +1,53 @@
...
+        # Short-term hack: just execute all tasks sequentially in the order of add_task. It's not
+        # optimal but the behavior is correct because child will only execute after all parents
+        # complete.
+        self._tasks.append(task)
+        return node
+
+    def run(self):
+        """The current implementation runs task sequentially in a thread pool."""
+        with ThreadPoolExecutor() as executor:
+            for task in self._tasks:
```
kevinkim9264, line 51: This execution currently relies on that the self._tasks list is in the order of parent -> children.
Can you add a TODO here on how we are going to execute by the DAG constructed by the parent -> children graph in the future?

kevinkim9264: Actually, shouldn't we execute it by nodes instead of tasks?

kevinkim9264 (changes requested):

pipeline/pipeline.py
yubing: Clarified offline and add more comments.

souravkhandelwal09: ideally, we can execute the nodes on the same level in parallel. But this will work for now

souravkhandelwal09 (approved):

pipeline/pipeline.py
yubing: Added comments to prevent users creating own Node instances.

kevinkim9264 (approved):

PR #77, opened 2021-11-03
Added checkpointing framework

souravkhandelwal09:
**Description**
1. Added checkpoint-service to carry out object level and pipeline level checkpointing.
2. Hooked up notebooks export/import, metastore export/import with the checkpoint framework

**Future work**
1. Hook up other objects with checkpoint framework
2. Add option to reset checkpoint

**Testing**
1. Added unit tests
2. Manually tested notebooks, metastore with and without `--use-checkpoint` flag.
```
python3 export_db.py --profile sourav-wm --download --use-checkpoint
python3 export_db.py --profile sourav-wm --metastore --use-checkpoint
python3 import_db.py --profile sourav-newWS-wm --metastore --use-checkpoint
```


dbclient/HiveClient.py
```diff
@@ -300,6 +302,27 @@ def export_hive_metastore(self, cluster_name=None, metastore_dir='metastore/', d
             print("Failed count: " + str(total_failed_entries))
             print("Total Databases attempted export: " + str(len(all_dbs)))
 
+    def get_successful_exported_metastore_tables(self, success_log):
+        tables = set()
```
mrchristine, line 306: Does this return tables + databases, what's the format we expect here? Fully qualified table names like `dbname.tablename`?

dbclient/HiveClient.py
```diff
@@ -300,6 +302,27 @@ def export_hive_metastore(self, cluster_name=None, metastore_dir='metastore/', d
...
+        if self.get_checkpoint_enabled() and os.path.exists(success_metastore_log_path):
+            print("Using checkpointing to load completed metastore tables")
+            with open(success_metastore_log_path, 'r') as fp:
+                for table in fp:
+                    item = json.loads(table.rstrip())
+                    tables.add(item['table'])
+        return tables
+
+    def get_successful_imported_metastore_tables(self, success_log):
+        tables = set()
```
mrchristine, line 317: Similar feedback for the exported table here. We should define whether this is just table name, dbname.tablename, or something else.

dbclient/WorkspaceClient.py
```diff
@@ -247,7 +248,7 @@ def download_notebooks(self, ws_log_file='user_workspace.log', ws_dir='artifacts
             for notebook_data in fp:
                 notebook_path = json.loads(notebook_data).get('path', None).rstrip('\n')
                 dl_resp = self.download_notebook_helper(notebook_path, export_dir=self.get_export_dir() + ws_dir)
-                if 'error_code' not in dl_resp:
+                if 'error' not in dl_resp:
```
mrchristine, line 251: If I recall, the response of `dl_resp` is a json object, so `error_code` is a key check for this json object. Changing to `error` may not produce the same result. Have we tested this error code and return path?

dbclient/WorkspaceClient.py
```diff
@@ -261,7 +262,9 @@ def download_notebook_helper(self, notebook_path, export_dir='artifacts/'):
         get_args = {'path': notebook_path, 'format': self.get_file_format()}
         if self.is_verbose():
             print("Downloading: {0}".format(get_args['path']))
-        resp = self.get(WS_EXPORT, get_args)
+        (checkpoint_available, resp) = self.checkpoint_get(WM_EXPORT, WORKSPACE_NOTEBOOK_OBJECT, notebook_path, WS_EXPORT, get_args)
```
mrchristine, line 265: Does `checkpoint_available` mean the notebook was in the checkpoint file? It's hard to understand what this means from the variable name.

dbclient/test/HiveClientTest.py
```diff
@@ -41,4 +41,36 @@ def mock_submit_cmd(cid, ec_id, repair_cmd):
             hiveClient.repair_legacy_tables(fix_table_log='test_repair_tables.log')
             output = fake_out.getvalue().split("\n")
             self.assertEqual(output[0], 'Table failed repair: default.test_legacy2')
-            self.assertEqual(output[1], '1 tables failed to repair. See errors in failed_repair_tables.log')
\ No newline at end of file
+            self.assertEqual(output[1], '1 tables failed to repair. See errors in failed_repair_tables.log')
```
mrchristine, line 44: Minor correction, `1 table(s)` should be singular.

dbclient/WorkspaceClient.py
souravkhandelwal09: dl_resp is returned from `download_notebook_helper` which is `error` as key in the json object. 
```
with open(self.get_export_dir() + 'failed_notebooks.log', 'a') as err_log:
            if resp.get('error', None):
                err_msg = {'error': resp.get('error'), 'path': notebook_path}
                err_log.write(json.dumps(err_msg) + '\n')
                return err_msg
```

dbclient/HiveClient.py
souravkhandelwal09: done

dbclient/WorkspaceClient.py
souravkhandelwal09: renamed it to `found_in_checkpoint`

dbclient/test/HiveClientTest.py
souravkhandelwal09: ✔️

dbclient/checkpoint_dbclient.py
```diff
@@ -0,0 +1,77 @@
...
+        if self._use_checkpoint:
+            checkpoint_file = self._get_object_checkpoint_file(action_type, object_type)
+            if self.is_completed_object(action_type, object_type, object_key):
+                logging.info(f"Object_type: {object_type}, object_key: {object_key} already exported")
+                return True, {}
+            else:
+                results = self.get(endpoint, json_params, version, print_json)
+                with open(checkpoint_file, 'a+') as append_fp:
+                    if not self._is_error(results):
+                        append_fp.write(object_key + "\n")
```
yubing, line 60: Another problem of checkpointing in dbclient: while exporting, the process could fail after this line and before the data is written into storage. It'll cause data loss.

Also, we need to look into how Python handles file caching to make sure we won't loss data in any scenarios.

dbclient/HiveClient.py
souravkhandelwal09: done

dbclient/checkpoint_dbclient.py
```diff
@@ -0,0 +1,77 @@
...
+        os.makedirs(self._checkpoint_dir, exist_ok=True)
+
+    def get_checkpoint_enabled(self):
+        return self._use_checkpoint
+
+    """
+    completed object_keys is populated from the export log of the objectType. This is used to start the export from
+    the last completed step.
+    """
+    def restore_checkpoint_objects(self, action_type, object_type):
```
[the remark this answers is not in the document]
souravkhandelwal09: as discussed offline, we don't want to do checkpointing for all objects and we want to give the flexibility to the clients to choose which objects to use checkpointing for

dbclient/checkpoint_dbclient.py
```diff
@@ -0,0 +1,77 @@
...
+                        self._completed_exported_objects.setdefault(object_type, set()).add(key.rstrip())
+                    else:
+                        self._completed_import_objects.setdefault(object_type, set()).add(key.rstrip())
+    """
+    Wrapper around dbclient.get() which does a conditional get based on request already checkpointed or not.
+    returns a tuple:
+    tuple_1 - denotes if the object_key is available from checkpoint (already exported)
+    tuple_2 - actual results from the api call. If tuple_1 is True, this value will be empty
+    """
+    def checkpoint_get(self, action_type, object_type, object_key, endpoint, json_params=None, version='2.0', print_json=False):
```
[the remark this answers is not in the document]
souravkhandelwal09: Changed the model by decoupling the checking point functionality and dbclient. It is more flexible now and can be used by the pipeline step checkpointing too.

dbclient/checkpoint_dbclient.py
souravkhandelwal09: Handled all instances of this happening. Added flushing of data wherever its needed

checkpoint_service.py
```diff
@@ -0,0 +1,71 @@
...
+        self._checkpoint_enabled = checkpoint_enabled
+        self._checkpoint_file = checkpoint_file
+        self._checkpoint_key_set = set()
+        self._restore_from_checkpoint_file()
+
+    # writes key to the checkpoint file
+    def write(self, key):
+        if self._checkpoint_enabled and key not in self._checkpoint_key_set:
+            with open(self._checkpoint_file, 'a+') as append_fp:
+                append_fp.write(key + "\n")
```
kevinkim9264, line 34: If `checkpoint_key_set` represents the set of keys that are checkpointed (written to the file), we also want to add the key to the _checkpoint_key_set?

checkpoint_service.py
```diff
@@ -0,0 +1,71 @@
...
+METASTORE_TABLES = "metastore"
+
+# Actions
+WM_EXPORT = "export"
+WM_IMPORT = "import"
+
+class CheckpointKeySet():
+    """Class that deals with checkpoint read and write of a particular object type"""
+
+    """
```
kevinkim9264, line 20: nit: No need to have a separate doc string for params? (may be we can merge them into one)

dbclient/WorkspaceClient.py
kevinkim9264: can we double check on this? How has this been working if `download_notebook_helper` indeed returns `error` as the key?

kevinkim9264 (approved):
Looks great to me! I guess we can technically further unify all the Client classes to have uniform write / read interface under which we can embed the checkpoint logics. But I think it is not worth it considering it will be quite time-consuming and it doesn't seem too hard to correctly embed the checkpoint logic for each of the object clients.

We just have to do the review well and unit test.

dbclient/HiveClient.py
```diff
@@ -365,14 +373,19 @@ def import_hive_metastore(self, cluster_name=None, metastore_dir='metastore/', v
...
+                    if checkpoint_metastore_set.contains(full_table_name):
+                        print(f"Table {full_table_name} found in checkpoint file, already imported")
                     else:
-                        print(f'Moving view ddl to re-apply later: {db_name}.{tbl_name}')
+                        print(f"Importing table {full_table_name}")
+                        local_table_ddl = metastore_local_dir + db_name + '/' + tbl_name
+                        if not self.move_table_view(db_name, tbl_name, local_table_ddl):
+                            # we hit a table ddl here, so we apply the ddl
+                            is_successful = self.apply_table_ddl(local_table_ddl, ec_id, cid, db_path, has_unicode)
+                            checkpoint_metastore_set.write(full_table_name)
```
yubing, line 385: Should we log the checkpoint when is_successful == false?

dbclient/HiveClient.py
```diff
@@ -426,23 +440,27 @@ def log_all_tables(self, db_name, cid, ec_id, metastore_dir, err_log_path, succe
...
                 tables_slice = 'print(all_tables[{0}:{1}])'.format(batch_size*m, batch_size*(m+1))
                 results = self.submit_command(cid, ec_id, tables_slice)
                 table_names = ast.literal_eval(results['data'])
                 for table_name in table_names:
                     print("Table: {0}".format(table_name))
-                    is_successful = self.log_table_ddl(cid, ec_id, db_name, table_name, metastore_dir,
-                                                       err_log_path, has_unicode)
+                    full_table_name = f'{db_name}.{table_name}'
+                    if checkpoint_metastore_set.contains(full_table_name):
+                        is_successful = 0
```
yubing, line 452: It's really weird assigning 0 to is_successful. Could you rename the variable or change the type to boolean to avoid the readability problem?

dbclient/HiveClient.py
```diff
@@ -426,23 +440,27 @@ def log_all_tables(self, db_name, cid, ec_id, metastore_dir, err_log_path, succe
...
                     if is_successful == 0:
-                        print(f"Exported {db_name}.{table_name}")
-                        success_item = {'table': f'{db_name}.{table_name}', 'iam': iam}
+                        print(f"Exported {full_table_name}")
+                        success_item = {'table': full_table_name, 'iam': iam}
                         sfp.write(json.dumps(success_item))
                         sfp.write('\n')
                     else:
                         print("Logging failure")
+                    checkpoint_metastore_set.write(full_table_name)
```
yubing, line 463: The same question. Should we checkpoint failed operations (which will prevent retry)? If this is expected, I think we should add a field to indicate success or failure in the checkpoint log.

I could imagine in addition to resume interrupted process, another major use case for checkpointing would be retry on failed operations.

dbclient/HiveClient.py
```diff
@@ -525,6 +543,7 @@ def retry_failed_metastore_export(self, cid, failed_metastore_log_path, iam_role
                             success_item = {'table': f'{db_name}.{table_name}', 'iam': iam_role}
                             sfp.write(json.dumps(success_item))
                             sfp.write('\n')
+                            checkpoint_metastore_set.write(f'{db_name}.{table_name}')
```
yubing, line 546: Ditto.

dbclient/HiveClient.py
yubing: This write is dangerous because the write to sfp in line #459 is not flushed yet. If the pipeline restarts after this point, data will be lost.

pipeline/pipeline_test.py
```diff
@@ -2,36 +2,78 @@
...
         pipeline.add_task(task_4, [node_2, node_3])
 
         pipeline.run()
 
-        self.assertEqual(AppendTask.result, [1, 2, 3, 4])
+        self.assertEqual(result, [1, 2, 3, 4])
+
+    def test_run_with_checkpoint(self):
+        result = []
+        task_1 = AppendTask("task1", 1, result)
```
yubing, line 44: Thanks for the testing. Could you help extract common logic of pipeline creation into a function?

tasks/tasks.py
```diff
@@ -63,3 +63,103 @@ def run(self):
         scim_c.import_all_users_and_groups()
         end = timer()
         logging.info("Complete Users and Groups Import Time: " + str(timedelta(seconds=end - start)))
+
+class WorkspaceItemLogTask(AbstractTask):
```
yubing, line 67: It would be great if we can have a separate PR for adding new tasks.

Both checkpointing and new tasks are very important, mixing them together will slow down both and we might overlook some issues.

yubing (changes requested):
As commented, it's not easy to get it right when enabling intra-task checkpointing for a component. So, I'd suggest:

1. From the framework's point of view, we should have a list of things to check in order to use it correctly and avoid common mistakes e.g. flushing of orders.
2. From the user's point of view, we should use it conservatively and carefully. The potential problems that could cause data loss are timing issues, which are really hard to catch with tests. Most realistic thing we can do is to review it carefully. So, I'd suggest to split framework, adoptions and other changes into separate PRs.

Thanks.

kevinkim9264:
> As commented, it's not easy to get it right when enabling intra-task checkpointing for a component. So, I'd suggest:
> 
> 1. From the framework's point of view, we should have a list of things to check in order to use it correctly and avoid common mistakes e.g. flushing of orders.
> 2. From the user's point of view, we should use it conservatively and carefully. The potential problems that could cause data loss are timing issues, which are really hard to catch with tests. Most realistic thing we can do is to review it carefully. So, I'd suggest to split framework, adoptions and other changes into separate PRs.
> 
> Thanks.

For #2, you mean the engineer who writes the checkpoint code in the task right? If so, I agree, but what is the case of data loss due to the timing issue? Because we are doing the checkpointing per key, would not skip getting the data for the not-checkpointed key (and we checkpoint only after the actual write). For the inter-tasks checkpointing, an issue like that can happen if the previous task it no completely done but marked as done, but even then I don't see a scenario where a task can be marked as checkpointed without every key being checkpointed.

dbclient/HiveClient.py
yubing: @kevinkim9264 let's take this as an example for the data loss risk I mentioned.

In line #459 and #460, data X is written into sfp but not flushed to underlying file yet because of buffering (https://docs.python.org/3/library/io.html#io.IOBase.flush). In line #463, the record has been logged into checkpoint file and flushed (see discussions in checkpoint_service.py).

If the process restarts right after #463, we will resume the previous progress according to checkpoint file. Then, we will skip logging data X because it exists in checkpoint file. But actually data X is never written into the log file which causes the data loss.

yubing:
> For #2, you mean the engineer who writes the checkpoint code in the task right? If so, I agree, but what is the case of data loss due to the timing issue? Because we are doing the checkpointing per key, would not skip getting the data for the not-checkpointed key (and we checkpoint only after the actual write). For the inter-tasks checkpointing, an issue like that can happen if the previous task it no completely done but marked as done, but even then I don't see a scenario where a task can be marked as checkpointed without every key being checkpointed.

Explained in the comments in HiveClient.py. For inter-task checkpointing, I'm less worried because there will be one single place to interactive with the checkpointing logic and won't change often, so it's much easier to validate.

tasks/tasks.py
souravkhandelwal09: One of main reason I added this WorkspaceItemTask as a part of this PR is to create visibility of how the checkpoint service can be used for pipeline task level checkpointing. (Since UserExport/ImportTask didn't require checkpointing).
TBH, the logic of adding new task is fairly straightforward, we are merely copying the logic from export_db.py and import_db.py. There is no new logic which is added at this point. 

what do you think ?

checkpoint_service.py
```diff
@@ -0,0 +1,71 @@
...
+    def __init__(self, checkpoint_enabled, checkpoint_file):
+        self._checkpoint_enabled = checkpoint_enabled
+        self._checkpoint_file = checkpoint_file
+        self._checkpoint_key_set = set()
+        self._restore_from_checkpoint_file()
+
+    # writes key to the checkpoint file
+    def write(self, key):
+        if self._checkpoint_enabled and key not in self._checkpoint_key_set:
+            with open(self._checkpoint_file, 'a+') as append_fp:
```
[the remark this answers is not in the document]
souravkhandelwal09: You are right that this will slow down performance but I was optimizing for correctness than performance in the first iteration of the checkpoint-service as the performance impact will be minimal. But since you brought it up, let me update the PR with the change. 

Also, do you think we need to flush the buffer after every write ? System crash is a rare occurrence which will cause data loss but we are fine with some amount of data loss in case of checkpoint keys. It will just redo the actual call for those keys. Doing flush after every write will affect performance.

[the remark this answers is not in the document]
yubing: I agree with that we should favor correctness over performance in this case. I'm ok to keep current implementation if you want.

On the other hand, I don't think any data loss is acceptable in WM, especially silent data loss.

tasks/tasks.py
yubing: I agree it makes sense to keep WorkspaceClient changes in this PR to demonstrate how checkpointing framework is used. Also I'm ok to keep checkpointing related changes in pipeline framework.

But adding WorkspaceItemTask is orthogonal to both intra-task checkpointing (changes in WorkspaceClient) and inter-task checkpointing (changes in pipeline framework). So I think it's better to exclude them from this PR.

checkpoint_service.py
souravkhandelwal09: _checkpoint_key_set is only populated at the start of the export / import of a particular object. Adding keys to the set in the middle doesn't add any value as we are never going to read that key

checkpoint_service.py
souravkhandelwal09: done

dbclient/HiveClient.py
souravkhandelwal09: Thanks for catching that, the checkpoint write needs to go inside the successful block. Fixed it and added flush to the sfp write. 

Yes, checkpoint service can be used for failure handling, and we can evolve the checkpoint service to handle that at a later point of time. The first iteration can be used just for resuming interruped process.

dbclient/HiveClient.py
souravkhandelwal09: added flush

dbclient/HiveClient.py
kevinkim9264: ah yeah, basically we need to make sure the write is "fully" done before we ever checkpoint. Once we ensure that, though, we should be safe from the data loss risk?

dbclient/WorkspaceClient.py
souravkhandelwal09: doubled checked the log file - `failed_notebooks.log`. The format is
```
{"error": "DatabricksServiceException: RESOURCE_DOES_NOT_EXIST: Path (/Users/ekl-backup-1/Intro Notebooks / Intro R Notebooks) doesn't exist.", "path": "/Users/ekl-backup-1/Intro Notebooks / Intro R Notebooks"}
```
I think it wasn't working before

tasks/tasks.py
souravkhandelwal09: moved them out of PR

dbclient/HiveClient.py
souravkhandelwal09: That's correct, all data to be flushed to disk before checkpointing happens

yubing: When we provide a feature or a library, in general, we should make it easy to use correctly and hard to use in a wrong way. However, the CheckpointService is opposite by design.

Even though we can use it correctly right now, changes in the future could easily break it, for example, someone thinks the flush is not necessary and remove it. On the other hand, making it perfect will take much longer, e.g. unify all file operations, a reasonable workaround is to over-documenting the potential risks in both library definition and all its references to make sure readers understand the risks.

@souravkhandelwal09 what do you think?

dbclient/HiveClient.py
yubing: Not changed?

dbclient/HiveClient.py
```diff
@@ -463,14 +482,14 @@ def log_table_ddl(self, cid, ec_id, db_name, table_name, metastore_dir, err_log_
             if ddl_str_resp['resultType'] != 'text':
                 ddl_str_resp['table'] = '{0}.{1}'.format(db_name, table_name)
                 err_log.write(json.dumps(ddl_str_resp) + '\n')
-                return -1
+                return False
```
yubing, line 485: Update comments accordingly.

dbclient/HiveClient.py
```diff
@@ -387,6 +400,7 @@ def import_hive_metastore(self, cluster_name=None, metastore_dir='metastore/', v
                     print("Importing view {0}.{1}".format(db_name, view_name))
                     local_view_ddl = metastore_view_dir + db_name + '/' + view_name
                     is_successful = self.apply_table_ddl(local_view_ddl, ec_id, cid, db_path, has_unicode)
+                    checkpoint_metastore_set.write(full_table_name)
```
yubing, line 403: This operation has been checkpointed but the checkpoint is not used?

dbclient/HiveClient.py
```diff
@@ -426,21 +440,26 @@ def log_all_tables(self, db_name, cid, ec_id, metastore_dir, err_log_path, succe
...
                 table_names = ast.literal_eval(results['data'])
                 for table_name in table_names:
                     print("Table: {0}".format(table_name))
-                    is_successful = self.log_table_ddl(cid, ec_id, db_name, table_name, metastore_dir,
-                                                       err_log_path, has_unicode)
-                    if is_successful == 0:
-                        print(f"Exported {db_name}.{table_name}")
-                        success_item = {'table': f'{db_name}.{table_name}', 'iam': iam}
+                    full_table_name = f'{db_name}.{table_name}'
+                    if checkpoint_metastore_set.contains(full_table_name):
```
yubing, line 451: I think this usage of checkpoint set is correct. But it's easily to be broken in future changes. Could you leave comments to explain why checkpointing is needed, how it works and the potential risks (e.g. why flush is necessary)? 

I know it's over-documenting, but I don't think the code is future proof in the current situation.

dbclient/WorkspaceClient.py
```diff
@@ -397,6 +408,8 @@ def log_acl_to_file(self, artifact_type, read_log_filename, write_log_filename,
                     continue
                 acl_resp.pop('http_status_code')
                 write_fp.write(json.dumps(acl_resp) + '\n')
+                write_fp.flush()
+                checkpoint_object_set.write(obj_id)
```
yubing, line 412: Ditto. Please over document :(

yubing (commented):
Thanks for making all the changes, Sourav!

Code mostly looks good to me. Please take a look the comments again. Thank you!

dbclient/HiveClient.py
souravkhandelwal09: What makes CheckpointService easy to use it the wrong way and hard to use it correctly ? Can you elaborate on that ?

Achieving pure fault tolerance will require providing a library of new file operations and overhaul of all the clients using it. That is going to be a massive effort. I kept the design such that its easy to use by the clients without disrupting much of its logic as we don't have test coverage in our clients. Also, I didn't want to spend a ton of time for something which is short lived, developed only by 3-4 committers, and doesn't have more use cases than ST migration. There is always a trade off and I think the risks are acceptable here. I am going to add some comments to highlight the risk of removing flush.

dbclient/HiveClient.py
souravkhandelwal09: I thought it would be handled by the checkpoint check in the forloop block before this as the views are only moved from there. But there would be scenario where we want to check for checkpoint here too. Good catch, added that

dbclient/HiveClient.py
souravkhandelwal09: Actually, apply_table_ddl function is just returning the response and was not handling errors, added to logic to checkpoint only on success

dbclient/HiveClient.py
yubing: Why I said it's easy to use it in the wrong way:
In order to use CheckpointService correctly, you'll need to keep in mind that the order of persistence i.e. flush (not the order of write) are always correct. Imagine if a 4th person who didn't review this PR is going to modify the code, how confident are you that they won't mess it up?

I totally agree with you that documenting is a reasonable workaround as I commented previously.

celcius.py
```diff
@@ -0,0 +1,19 @@
+# Using @property decorator
```
yubing, line 1: Testing code?

dbclient/HiveClient.py
```diff
@@ -426,21 +448,28 @@ def log_all_tables(self, db_name, cid, ec_id, metastore_dir, err_log_path, succe
...
+                        is_successful = True
```
yubing, line 460: print some info about checkpoint hit?

dbclient/HiveClient.py
yubing: This checkpoint is written but not used?

dbclient/WorkspaceClient.py
```diff
@@ -389,6 +398,8 @@ def log_acl_to_file(self, artifact_type, read_log_filename, write_log_filename,
             for x in read_fp:
                 data = json.loads(x)
                 obj_id = data.get('object_id', None)
+                if checkpoint_object_set.contains(obj_id):
```
yubing, line 401: Print or log something.

yubing (approved):
LGTM with some nits.

Thanks you for make rounds of changes to mitigate the risks! Really appreciate it.

checkpoint_service.py
```diff
@@ -0,0 +1,106 @@
...
+        self._checkpoint_file = checkpoint_file
+        self._checkpoint_file_append_fp = open(checkpoint_file, 'a+')
+        self._checkpoint_key_set = set()
+        self._restore_from_checkpoint_file()
+
+    def write(self, key):
+        """Writes key into checkpoint file. This also flushes data after write to prevent loss of checkpoint data on
+        system crash.
+
+        Note: Make sure to persist your data before calling write. There is risk of data loss if your data is not persisted
```
yubing, line 51: Nit: could you make it more alarming? Something like !!!CAUTION!!!

celcius.py
souravkhandelwal09: removed

dbclient/HiveClient.py
souravkhandelwal09: it's getting logged in CheckpointKeySet.contains

dbclient/WorkspaceClient.py
souravkhandelwal09: same here

checkpoint_service.py
souravkhandelwal09: done

dbclient/WorkspaceClient.py
```diff
@@ -397,6 +408,8 @@ def log_acl_to_file(self, artifact_type, read_log_filename, write_log_filename,
                     continue
                 acl_resp.pop('http_status_code')
                 write_fp.write(json.dumps(acl_resp) + '\n')
+                write_fp.flush()
```
kevinkim9264, line 411: 2 questions:
1. I am worried if we do the flushing on every key, the performance can be visibly degraded. Is this something we should worry about?
2. Since the flush will just make sure flushing the python internal buffered data to the os buffered data, are we stating that we are just guarding against the Python app crash scenario, rather than the whole OS system crashing?

kevinkim9264, line 411: 2 questions:

1. I am worried if we do the flushing on every key, the performance can be visibly degraded. Is this something we should worry about?

2. Since the flush will just make sure flushing the python internal buffered data to the os buffered data, are we stating that we are just guarding against the Python app crash scenario, rather than the whole OS system crashing?

kevinkim9264 (commented):
one comment

dbclient/WorkspaceClient.py
souravkhandelwal09: Results from running jobs:
**Original**
Complete Workspace Export Time: 0:00:02.802726
Complete Notebook ACLs Export Time: 0:00:04.213943
Complete Directories ACLs Export Time: 0:00:02.950634

**New**
Complete Workspace Export Time: 0:00:02.884954 (negligible change)
Complete Notebook ACLs Export Time: 0:00:04.948932 (+15%)
Complete Directories ACLs Export Time: 0:00:03.211578 (+10%)

souravkhandelwal09: @kevinkim9264 the performance is minimally impacted with the flush

nfx (changes requested):
Please fix the errors from automated code review: https://lgtm.com/projects/g/databrickslabs/migrate/rev/pr-9382ed2d4f2b77ca7c5e3e709abc111d29752ff2

dbclient/WorkspaceClient.py
yubing: Re Kevin:

1. I think the performance degradation due to flushing is eligible because we only choose to checkpoint slow operations which should take at least dozens of seconds or minutes.
2. It's a good point. I think we should call os.fsync to guarantee the correctness in any scenarios. We can wrap flush and fsync in one function, so there is no extra complexity. Also, performance penalty should be acceptable because disk sync takes ~10ms according to https://gist.github.com/jboner/2841832.

kevinkim9264: @souravkhandelwal09 thank you for the benchmark result report!

@yubing 
1. I assume you meant negligible. And I agree that it's negligible if we indeed choose to checkpoint only for the heavy operations and just do the checkpoint at the stage level for the small operation ones. 
2. Yeah, though the chance is super super small for OS to crash and the checkpoint disk write to happen before the data save happen, if we are targeting 100% confidence level, we should consider fsyncing. With that said, this is only eligible if we are indeed checkpointing only for the heavy operations. If we fsync on every key operations, I believe the performance degradation will be an issue.

souravkhandelwal09:
> Please fix the errors from automated code review: https://lgtm.com/projects/g/databrickslabs/migrate/rev/pr-9382ed2d4f2b77ca7c5e3e709abc111d29752ff2

Fixed all issues except the pattern `from X import *`. This pattern is used at other places too and I believe we can live with this pattern

nfx:
@souravkhandelwal09 star imports are considered as a bad practice, as they may also overwrite the current namespace. Even if it's used in 5 other places, it doesn't mean it should be fine to keep using it :)

dbclient/WorkspaceClient.py
kevinkim9264: got it

souravkhandelwal09:
> @souravkhandelwal09 star imports are considered as a bad practice, as they may also overwrite the current namespace. Even if it's used in 5 other places, it doesn't mean it should be fine to keep using it :)

@nfx all alerts are fixed now

PR #80, opened 2021-11-17
Log necessary files to support EditClusterOwner

kevinkim9264:
Currently, all the clusters that are imported have their "creator" attribute to point to the owner of the PAT token used to do the import. i.e. if `abcd@databricks.com`'s PAT token is used to import the clusters, all the imported clusters' Creator attributes are set to `abcd@databricks.com` which can be different from the original creator of the clusters in the source workspace.

In order to help giving the correct creator attribute for the migrated clusters, in this PR, we output some log files which can be handed to Databricks teams to edit the migrated cluster's creator attribute accordingly.

The PR makes the following change
`python3 import_db.py --profile xxx --users` outputs:
- `user_name_to_user_id.log` which consists of username to userId mapping of the destination workspace after the user import is done. This file is used later to connect the original cluster creator username to the userId in the destination workspace.

`python3 import_db.py --profile xxx --clusters` outputs:
- `original_creator_user_ids.log` which consists of the list of userIds of the original creators of the imported clusters.
- `cluster_ids_to_change_creator.log` which consists of the list of clusterIds in the destination workspace to change their creator attributes.

The invoker of the script will give `original_creator_user_ids.log` and `cluster_ids_to_change_creator.log` to the Databricks side who will run a Jenkins job to do the cluster owner edits.

Example outputs:
`original_creator_user_ids.log`
```
new_cluster_id_1","new_cluster_id_2","new_cluster_id_3"
```

`cluster_ids_to_change_creator.log`
```
"original_cluster_creator_id_1","original_cluster_creator_id_2","original_cluster_creator_id_3"
```

**Unit Test:** 
```
python3 -m unittest dbclient/test/ClustersClientTest.py
```

**Manul Test**: 
Used the `original_creator_user_ids.log` and `cluster_ids_to_change_creator.log` files directly to the Databricks tool to edit the cluster owner changes.


dbclient/ClustersClient.py
```diff
@@ -277,6 +281,59 @@ def import_cluster_configs(self, log_file='clusters.log', acl_log_file='acl_clus
...
+                current_cluster_id, current_cluster_creator = current_cluster_name_to_id_and_creator_name[cluster_name]
+
+                # Log only if the cluster's current creator is not the same as the original creator
+                if original_cluster_creator != current_cluster_creator:
+                    if cluster_name in current_cluster_name_to_id_and_creator_name and \
+                            original_cluster_creator in user_name_to_user_id:
+                        cluster_ids_to_change_creator.append(current_cluster_id)
+                        original_creator_user_ids.append(user_name_to_user_id[original_cluster_creator])
+
+        with open(self.get_export_dir() + cluster_ids_file, 'w') as fp:
```
yubing, line 331: Just curious, no change required. 

Why do we output two separate files instead of writing related data into the same file? Copy-paste twice is more error-prone than once.

dbclient/test/ClustersClientTest.py
```diff
@@ -0,0 +1,63 @@
...
+            {
+                'cluster_name': "cluster_1",
+                'cluster_id': 'new_cluster_id_1',
+                'creator_user_name': 'new_cluster_creator_user_name_1'
+            },
+            {
+                'cluster_name': "cluster_2",
+                'cluster_id': 'new_cluster_id_2',
+                'creator_user_name': 'new_cluster_creator_user_name_2'
+            },
```
yubing, line 31: Can you add coverage for the case that multiple clusters with the same name?

dbclient/test/ClustersClientTest.py
```diff
@@ -0,0 +1,63 @@
...
+            {
+                'cluster_name': "cluster_3",
```
yubing, line 33: Also coverage for the current creator is correctly set?

yubing (approved):
LGTM with some nits.

dbclient/ClustersClient.py
kevinkim9264: It is because the Databricks side tool (the jenkins job for editing cluster owner) takes two separate parameters which are list. From the Jenkins job perspective, two-lists is the least error-prone way as other types of inputs are much easier to make mistakes on. (e.g. Json blobs, or list of tuples, etc) 

From this copy-pasting perspective, it is to reduce the chance of making mistakes on the Databricks side:
the invoker just gives these two files to Databricks, and Databricks directly copies pastes. Another similar way would be to make this in to one file, and let Databricks do the careful copy, but I think the current way is less error-prone.

souravkhandelwal09 (commented):
minor comments, LGTM overall

dbclient/test/ClustersClientTest.py
kevinkim9264: √

dbclient/test/ClustersClientTest.py
kevinkim9264: yeah, now that I changed the logic a little bit, it will always output even if the creator doesn't change. 
(because barely will there be any case where the new creator is the same as the original creator)

kevinkim9264: But let me still add the case

dbclient/ClustersClient.py
```diff
@@ -277,6 +281,59 @@ def import_cluster_configs(self, log_file='clusters.log', acl_log_file='acl_clus
...
+            else:
+                current_cluster_name_to_id_and_creator_name[cluster_name] = \
+                    (cluster['cluster_id'], cluster['creator_user_name'])
+
+        with open(self.get_export_dir() + cluster_log_file, 'r') as fp:
+            for line in fp:
+                cluster_conf = json.loads(line)
+                cluster_name = cluster_conf['cluster_name']
+                original_cluster_creator = cluster_conf['creator_user_name']
+                current_cluster_id, current_cluster_creator = current_cluster_name_to_id_and_creator_name[cluster_name]
```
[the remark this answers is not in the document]
kevinkim9264: that logic is gone now

dbclient/ClustersClient.py
```diff
@@ -277,6 +281,59 @@ def import_cluster_configs(self, log_file='clusters.log', acl_log_file='acl_clus
...
+
+                # Log only if the cluster's current creator is not the same as the original creator
+                if original_cluster_creator != current_cluster_creator:
+                    if cluster_name in current_cluster_name_to_id_and_creator_name and \
+                            original_cluster_creator in user_name_to_user_id:
```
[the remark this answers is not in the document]
kevinkim9264: √

dbclient/ClustersClient.py
```diff
@@ -277,6 +281,59 @@ def import_cluster_configs(self, log_file='clusters.log', acl_log_file='acl_clus
...
+            cluster_log_file,
+            user_name_to_user_id_log_file='user_name_to_user_id.log',
+            creators_file='original_creator_user_ids.log',
+            cluster_ids_file='cluster_ids_to_change_creator.log'):
+        """
+        Log the cluster_ids_to_change_creator and original_creator_user_ids
+        These can be used to edit the clusters to be owned by the correct/original creator instead of the
+        PAT token owner. Fails if user_name_to_user_id.log does not exist.
+
+        :param cluster_log_file: file that contains the exported cluster objects.
```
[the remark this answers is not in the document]
kevinkim9264: √

dbclient/JobsClient.py
```diff
@@ -195,26 +195,3 @@ def delete_all_jobs(self):
         job_list = self.get('/jobs/list').get('jobs', [])
         for job in job_list:
             self.post('/jobs/delete', {'job_id': job['job_id']})
-
-    def get_cluster_id_mapping(self, log_file='clusters.log'):
```
kevinkim9264, line 199: This is moved to ClustersClient.py

dbclient/ClustersClient.py
```diff
@@ -277,6 +281,74 @@ def import_cluster_configs(self, log_file='clusters.log', acl_log_file='acl_clus
...
+                    print("The old cluster id " + original_cluster_id +
+                          " does not get logged for EditClusterOwner due to some problems.")
+
+        with open(self.get_export_dir() + cluster_ids_file, 'w') as fp:
+            fp.write(json.dumps(cluster_ids_to_change_creator, separators=(',', ':')))
+
+        with open(self.get_export_dir() + creators_file, 'w') as fp:
+            fp.write(json.dumps(original_creator_user_ids, separators=(',', ':')))
+
+    def get_cluster_id_mapping(self, log_file='clusters.log'):
```
kevinkim9264, line 328: moved from JobsClient.py.

dbclient/test/ClustersClientTest.py
kevinkim9264: done

souravkhandelwal09 (approved):

dbclient/ClustersClient.py
```diff
@@ -277,6 +281,77 @@ def import_cluster_configs(self, log_file='clusters.log', acl_log_file='acl_clus
...
+            for line in fp:
+                cluster_conf = json.loads(line)
+                original_cluster_creator = cluster_conf['creator_user_name']
+                original_cluster_id = cluster_conf['cluster_id']
+                if original_cluster_id in old_to_new_cluster_mapping and original_cluster_creator in user_name_to_user_id:
+                    current_cluster_id = old_to_new_cluster_mapping[original_cluster_id]
+                    cluster_ids_to_change_creator.append(current_cluster_id)
+                    original_creator_user_ids.append(user_name_to_user_id[original_cluster_creator])
+                else:
+                    print("The old cluster id " + original_cluster_id +
```
yubing, line 319: Nit: we should also log original_cluster_creator in the error message.

yubing (approved):
Approved with a nit.

dbclient/ClustersClient.py
kevinkim9264: √

dbclient/ClustersClient.py
mrchristine: I agree, as long as we're providing the downloaded files directly, it should not be error prone. If we are copying / pasting values, that would be error prone and we shouldn't manually modify these values.

mrchristine (approved):
lgtm!

dbclient/ClustersClient.py
kevinkim9264: yeah, whoever runs the Jenkins job needs to do the copying pasting, but no need to change values

kevinkim9264:
jenkins merge

PR #82, opened 2021-12-02
Implement json_diff library for data integrity verification.

yubing:
This PR implements a framework to compare two json files for data integrity verification of workspace migration pipeline.

3 parts are planned as below:
- [x] A library to compare to Python dict which only contains dict and prime data types.
- [x] A library to covert JSON to Python dict per configuration which indicates primary key of a dict in order to convert lists to dict.
- [x] Support ID mapping. Basically, ignore certain fields.

Tested by:
``` python3 -m unittest validate/json_diff_test.py```
And
```
 python3 migration_pipeline.py --validate-pipeline --cluster-name='dummy' --source-session=20211207104023 --destination-session=20211207142122

|aaron@databricks.com|displayName:MISS_LEFT:
> aaron@databricks.com

|aaron@databricks.com|name:MISS_LEFT:
> {'givenName': 'aaron@databricks.com'}

|adrian@databricks.com|displayName:MISS_LEFT:
> adrian@databricks.com

|adrian@databricks.com|name:MISS_LEFT:
> {'givenName': 'adrian@databricks.com'}

|adriana.tufa@databricks.com|displayName:MISS_LEFT:
> adriana.tufa@databricks.com

|adriana.tufa@databricks.com|name:MISS_LEFT:
> {'givenName': 'adriana.tufa@databricks.com'}
```

### CODE REVIEW

**Comment by souravkhandelwal09** [comment]
I have a few concerns/questions with this approach:
1. The library is doing in-memory data comparison. It may need a lot of memory to work for large customers which has large number of notebooks. I am guessing that this job is going to be run from the devbox. Our current recommendation for devbox instanceType is `t3.micro` which has 1GB of memory. This memory will be insufficient for this job. Can you verify how much memory will be needed based on the largest customer's data ?(I think @kevinkim9264 had a spreadsheet which has the distribution of notebooks data across customers)
2. Constructing the `PrimaryKeyConfig` may be cumbersome but given that we have only fixed number of ObjectTypes we are dealing with, it should be fine as a one time effort. However, any change in data format will break this and we need to redo the `PrimaryKeyConfig` again.

One way to simplify this and not have to worry about memory constraints would be to leverage databricks notebooks to do this. The way that would work:
1. upload exported data from ST to destination DBFS
2. export the imported data in the same DBFS. 
3. These files can be read as spark DataFrames without much effort or translation logic. 
4. Finding diffs in DFs is straighforward and results can be published in a more readable format. 

WDYT ?

File: validate/json_diff.py [hunk 1]
```diff
@@ -0,0 +1,169 @@
...
+    if type(left) != type(right):
+        return TypeDiff(left, right)
+
+    if isinstance(left, (int, float, str)):
+        if left != right:
+            return ValueDiff(left, right)
+    elif isinstance(left, dict):
+        diff = DictDiff()
+        for key in set(left.keys()).union(set(right.keys())):
+            if key not in left:
```
kevinkim9264, line 102: just a note: If the export of the destination workspace worked properly, this case (or the other case) just means the destination workspace has extra data, which can happen if the workspace was being used beforehand. We may want to just warn in that case, though we can't do that here since we don't know whether left or right is the source vs dst.

validate/json_diff.py
```diff
@@ -0,0 +1,169 @@
...
+    children: dict = field(default_factory=dict)
+
+
+def prepare_diff_input(data, config=None):
+    """
+    1) List of primary types to sets.
+    2) List of dict to dict of dicts keyed by primary key.
+
+    :param data:
+    :param config:
```
kevinkim9264, line 139: Please fill out the param doc. (especially on the format of config)

kevinkim9264 (commented):
Not so much comments on the codes. However, can we discuss more on how this framework will be used end-to-end?

For example, 1. read object files from logs/ dir 2. load one object type data 3. do the prepare_diff_input, etc.

Then we can see some of the gaps we will need to fill out such as  streaming or dynamic sampling necessity since all the data (even if it's just one object type) may not fit in memory  

Also we need to touch the pipeline a little bit to be able to export the destination workspace and save the log files accordingly.

kevinkim9264:
> I have a few concerns/questions with this approach:
> 
> 1. The library is doing in-memory data comparison. It may need a lot of memory to work for large customers which has large number of notebooks. I am guessing that this job is going to be run from the devbox. Our current recommendation for devbox instanceType is `t3.micro` which has 1GB of memory. This memory will be insufficient for this job. Can you verify how much memory will be needed based on the largest customer's data ?(I think @kevinkim9264 had a spreadsheet which has the distribution of notebooks data across customers)


The devbox creation is done by the users and hence the size can change for sure. Another way to deal with memory issue is to sampling but it is not easy since we would still have to sample by the keys (but this can be done by streaming from disks)


> One way to simplify this and not have to worry about memory constraints would be to leverage databricks notebooks to do this. The way that would work:
> 
> 1. upload exported data from ST to destination DBFS
> 2. export the imported data in the same DBFS.
> 3. These files can be read as spark DataFrames without much effort or translation logic.
> 4. Finding diffs in DFs is straighforward and results can be published in a more readable format.
> 
> WDYT ?

I agree that using the Spark can potentially simplify the work and give more promising scalability. It is feasible only if we can use the dbfs APIs to manage the full life-cycle of this work: upload all the exported data and delete the data after doing the comparisons. With that said, I think it will be a bit more work, and also more intrusive, because the Field eng / invokers will likely need more interactions with the customer dst workspace (read the diffs files, etc)

yubing:
> I have a few concerns/questions with this approach:
> 
> 1. The library is doing in-memory data comparison. It may need a lot of memory to work for large customers which has large number of notebooks. I am guessing that this job is going to be run from the devbox. Our current recommendation for devbox instanceType is `t3.micro` which has 1GB of memory. This memory will be insufficient for this job. Can you verify how much memory will be needed based on the largest customer's data ?(I think @kevinkim9264 had a spreadsheet which has the distribution of notebooks data across customers)

Thanks for pointing it out. You are right this library is purely in-memory. But I'm not very concerned by the size of object files because it's straightforward to shard and sample JSON files.

> 2. Constructing the `PrimaryKeyConfig` may be cumbersome but given that we have only fixed number of ObjectTypes we are dealing with, it should be fine as a one time effort. However, any change in data format will break this and we need to redo the `PrimaryKeyConfig` again.

You are right that PrimaryKeyConfig should be consistent with output file format. But given the output file format is mostly determined by public API, which shouldn't change often.

> 
> One way to simplify this and not have to worry about memory constraints would be to leverage databricks notebooks to do this. The way that would work:
> 
> 1. upload exported data from ST to destination DBFS
> 2. export the imported data in the same DBFS.
> 3. These files can be read as spark DataFrames without much effort or translation logic.
> 4. Finding diffs in DFs is straighforward and results can be published in a more readable format.
> 
> WDYT ?

I didn't choose Spark or other tools because 1) minimize dependencies; 2) make results easy to be parsed by human.

After translating JSON files to DFs, it will be harder for human to make sense of the diffs in my opinion.

validate/json_diff.py
yubing: Done.

migration_pipeline.py
```diff
@@ -88,15 +102,67 @@ def build_import_pipeline(client_config, checkpoint_service, args) -> Pipeline:
...
+    def add_diff_task(name, config, file_path, parents=None):
+        source_file = os.path.join(source_dir, file_path)
+        destination_file = os.path.join(destination_dir, file_path)
+        return pipeline.add_task(DiffTask(name, source_file, destination_file, config), parents)
+
+    add_diff_task(
+        "validate-users",
+        DiffConfig(
+            primary_key='userName',
+            ignored_keys={'id', 'entitlements'},
```
souravkhandelwal09, line 141: why ignore entitlements ?

validate/json_diff.py
```diff
@@ -0,0 +1,203 @@
...
+    else:
+        raise NotImplementedError(f"Type {type(left)} is not supported.")
+
+    return None
+
+
+class DiffConfig:
+    """
+    DiffConfig configures how to diff two Python dicts. It should have the same structure as `data`,
+    and follow the the rules below:
```
souravkhandelwal09, line 132: nit: `the the` -> the

souravkhandelwal09 (approved):
LGTM with minor nits

migration_pipeline.py
kevinkim9264, line 141: can you leave comments on why entitlements is also ignored?

kevinkim9264 (changes requested):
looks good. some nits

validate/json_diff.py
yubing: Per offline discussions, the output of this tool is to help users understand the diff. It doesn't do anything more than warning.

migration_pipeline.py
yubing: Good catch. entitlements shouldn't be ignored. It was for testing and let me remove it.

validate/json_diff.py
yubing: Done.

dbclient/parser.py
```diff
@@ -428,6 +436,15 @@ def get_pipeline_parser() -> argparse.ArgumentParser:
     parser.add_argument('--import-pipeline', action='store_true',
                         help='Execute all import tasks.')
 
+    parser.add_argument('--validate-pipeline', action='store_true',
+                        help='Validate exported data between source and destination.')
+
+    parser.add_argument('--source-session', action='store', default='',
```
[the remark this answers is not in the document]
yubing: Done.

migration_pipeline.py
```diff
@@ -5,29 +5,35 @@
...
+
+        # Cant use netrc credentials because requests module tries to load the credentials into http
+        # basic auth headers parse the credentials
+        url = login_args['host']
+        token = login_args['token']
+        client_config = parser.build_client_config(args.profile, url, token, args)
+
     client_config['session'] = session
-    client_config['export_dir'] = os.path.join(client_config['export_dir'], session) + '/'
+    client_config['base_dir'] = client_config['export_dir']
```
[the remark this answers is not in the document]
yubing: Done.

migration_pipeline.py
yubing: Reverted.

kevinkim9264 (approved):

PR #83, opened 2021-12-06
Make it work for dbadmin user in all cases

kevinkim9264:
It turned out that not all workspaces (for legacy or w.e. reasons) necessarily have the "admin" folders. So if the invoker is dbadmin user, we create the tmp directory on the top level directory


souravkhandelwal09 (approved):

mrchristine (approved):
looks good to me. The /tmp/ top level path will only have admin access and where we can store these tmp tools.

PR #84, opened 2021-12-07
Added error handling to output errors in failed log files

souravkhandelwal09:
**Description**
1. More consistent way of logging failures - the failed logs will appear in `failed_{action_type}_{import_type}.log` files.
2. Logging errors from all the places in WorkspaceClient.py (Other clients to follow in subsequent PRs)
3. migration-pipeline will stop if failed logs have errors. 
4. Fixed Notebook ACL import to skip user directory ACL import for non existent users.

**Manual Test**
Export - ```python migration-pipeline.py --profile e2-demo-src --export-pipeline --use-checkpoint --cluster-name API_Metastore_Work_Leave_Me_Alone```

Import - ```--profile e2-demo-dst --import-pipeline --use-checkpoint --repair-metastore-tables --session 20211206151424 --cluster-name API_Metastore_Work_Leave_Me_Alone --archive-missing```

- During import of notebook acls, some errors were encountered (relating to path not existent). Trying to find more info on why those paths were not created.

### DIFF
--- a/dbclient/WorkspaceClient.py
+++ b/dbclient/WorkspaceClient.py
@@ -3,6 +3,8 @@ from dbclient import *
 import wmconstants
 from timeit import default_timer as timer
 from datetime import timedelta
+import logging_utils
+import logging
 import os
 
 WS_LIST = "/workspace/list"
@@ -245,6 +247,8 @@ class WorkspaceClient(dbclient):
         checkpoint_notebook_set = self._checkpoint_service.get_checkpoint_key_set(
             wmconstants.WM_EXPORT, wmconstants.WORKSPACE_NOTEBOOK_OBJECT)
         ws_log = self.get_export_dir() + ws_log_file
+        notebook_error_logger = logging_utils.get_error_logger(
+            wmconstants.WM_EXPORT, wmconstants.WORKSPACE_NOTEBOOK_OBJECT, self.get_export_dir())
         num_notebooks = 0
         if not os.path.exists(ws_log):
             raise Exception("Run --workspace first to download full log of all notebooks.")
@@ -253,12 +257,13 @@ class WorkspaceClient(dbclient):
             # pull the path from the data to download the individual notebook contents
             for notebook_data in fp:
                 notebook_path = json.loads(notebook_data).get('path', None).rstrip('\n')
-                dl_resp = self.download_notebook_helper(notebook_path, checkpoint_notebook_set, export_dir=self.get_export_dir() + ws_dir)
+                dl_resp = self.download_notebook_helper(notebook_path, checkpoint_notebook_set, notebook_error_logger,
+                                                        export_dir=self.get_export_dir() + ws_dir)
                 if 'error' not in dl_resp:
                     num_notebooks += 1
         return num_notebooks
 
-    def download_notebook_helper(self, notebook_path, checkpoint_notebook_set, export_dir='artifacts/'):
+    def download_notebook_helper(self, notebook_path, checkpoint_notebook_set, error_logger, export_dir='artifacts/'):
         """
         Helper function to download an individual notebook, or log the failure in the failure logfile
         :param notebook_path: an individual notebook path
@@ -269,13 +274,12 @@ class WorkspaceClient(dbclient):
             return {'path': notebook_path}
         get_args = {'path': notebook_path, 'format': self.get_file_format()}
         if self.is_verbose():
-            print("Downloading: {0}".format(get_args['path']))
+            logging.info("Downloading: {0}".format(get_args['path']))
         resp = self.get(WS_EXPORT, get_args)
-        with open(self.get_export_dir() + 'failed_notebooks.log', 'a') as err_log:
-            if resp.get('error', None):
-                err_msg = {'error': resp.get('error'), 'path': notebook_path}
-                err_log.write(json.dumps(err_msg) + '\n')
-                return err_msg
+        if resp.get('error', None):
+            err_msg = {'error': resp.get('error'), 'path': notebook_path}
+            error_logger.error(json.dumps(err_msg) + '\n')
+            return err_msg
         nb_path = os.path.dirname(notebook_path)
         if nb_path != '/':
             # path is NOT empty, remove the trailing slash from export_dir
@@ -347,7 +351,7 @@ class WorkspaceClient(dbclient):
         items = self.get(WS_LIST, get_args).get('objects', None)
         num_nbs = 0
         if self.is_verbose():
-            print("Listing: {0}".format(get_args['path']))
+            logging.info("Listing: {0}".format(get_args['path']))
         if items is not None:
             # list all the users folders only
             folders = self.filter_workspace_items(items, 'DIRECTORY')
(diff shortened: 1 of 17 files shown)


kevinkim9264:
@souravkhandelwal09  on 
```
During import of notebook acls, some errors were encountered (relating to path not existent). Trying to find more info on why those paths were not created.
```
could you find the error logs? and put it into [go/wmscriptsop (error handling section)](https://docs.google.com/document/d/1SX7yB5Ti_mPZ-eLWlZYQvwOxluyScMjhpsyMvSJfZGk/edit#heading=h.mmap4fyfjbt5)?

dbclient/WorkspaceClient.py
```diff
@@ -245,6 +247,8 @@ def download_notebooks(self, ws_log_file='user_workspace.log', ws_dir='artifacts
         checkpoint_notebook_set = self._checkpoint_service.get_checkpoint_key_set(
             wmconstants.WM_EXPORT, wmconstants.WORKSPACE_NOTEBOOK_OBJECT)
         ws_log = self.get_export_dir() + ws_log_file
+        notebook_error_logger = logging_utils.get_error_logger(
+            wmconstants.WM_EXPORT, wmconstants.WORKSPACE_NOTEBOOK_OBJECT, self.get_export_dir())
```
yubing, line 251: I think it's better to organize all error logs into one dir?

dbclient/WorkspaceClient.py
```diff
@@ -253,12 +257,13 @@ def download_notebooks(self, ws_log_file='user_workspace.log', ws_dir='artifacts
...
                 notebook_path = json.loads(notebook_data).get('path', None).rstrip('\n')
-                dl_resp = self.download_notebook_helper(notebook_path, checkpoint_notebook_set, export_dir=self.get_export_dir() + ws_dir)
+                dl_resp = self.download_notebook_helper(notebook_path, checkpoint_notebook_set, notebook_error_logger,
+                                                        export_dir=self.get_export_dir() + ws_dir)
                 if 'error' not in dl_resp:
                     num_notebooks += 1
         return num_notebooks
 
-    def download_notebook_helper(self, notebook_path, checkpoint_notebook_set, export_dir='artifacts/'):
+    def download_notebook_helper(self, notebook_path, checkpoint_notebook_set, error_logger, export_dir='artifacts/'):
```
yubing, line 266: I think it's a common pattern to define logger as glob variable to void passing it around.

dbclient/WorkspaceClient.py
```diff
@@ -439,24 +446,33 @@ def apply_acl_on_object(self, acl_str):
...
+            ws_user = self.get_user(obj_path)
+            if not self.does_user_exist(ws_user):
+                logging.info(f"User workspace does not exist: {obj_path}, skipping ACL")
+                return
         obj_status = self.get(WS_STATUS, {'path': obj_path})
+        if 'error_code' in obj_status:
+            error_logger.error(json.dumps(obj_status) + '\n')
+            return
         print("ws-stat: ", obj_status)
         current_obj_id = obj_status.get('object_id', None)
```
yubing, line 459: logging.info

export_db.py
```diff
@@ -28,6 +28,7 @@ def main():
     client_config = build_client_config(args.profile, url, token, args)
 
     os.makedirs(client_config['export_dir'], exist_ok=True)
+    logging_utils.set_default_logging(client_config['export_dir'])
```
yubing, line 31: Can we have a dir for logging?

import_db.py
```diff
@@ -22,6 +23,7 @@ def main():
     client_config = build_client_config(args.profile, url, token, args)
 
     makedirs(client_config['export_dir'], exist_ok=True)
+    logging_utils.set_default_logging(client_config['export_dir'])
```
yubing, line 26: Ditto.

migration_pipeline.py
```diff
@@ -29,12 +30,14 @@ def build_pipeline(args) -> Pipeline:
     client_config['session'] = session
     client_config['export_dir'] = os.path.join(client_config['export_dir'], session) + '/'
 
-    if client_config['debug']:
-        logging.info(url, token)
-
     if not args.dry_run:
         os.makedirs(client_config['export_dir'], exist_ok=True)
 
+    logging_utils.set_default_logging(client_config['export_dir'])
```
yubing, line 36: Ditto.

dbclient/dbclient.py
```diff
@@ -6,6 +6,7 @@
 from dbclient import parser
 import time
 import requests.packages.urllib3
+import logging
```
yubing, line 9: Can we add common error logging in dbclient which shared by all clients?

It'll release our burden for adding error logging in each of the clients. We can only add extra logging for important errors.

yubing (approved):
LGTM with some nits.

dbclient/WorkspaceClient.py
kevinkim9264: +1. It will be easier for the invokers to find.

pipeline/pipeline.py
```diff
@@ -71,4 +73,10 @@ def _run_task(self, task: AbstractTask):
             task.run()
         end = timer()
         logging.info(f'{task.name} Completed. Total time taken: {str(timedelta(seconds=end - start))}')
+        failed_task_log = logging_utils.get_error_log_file(task.action_type, task.object_type, self._working_dir)
+        # TODO: Add logic to skip failure check
```
kevinkim9264, line 77: I think the pipeline should always fail if there are some failures. If the invoker for sure knows that he/she can skip the failure, then the user can simply skip that task via --skip-tasks.

btw, can we make sure the task is not checkpointed if the pipeline throws due to the error log files?

kevinkim9264 (approved):
Thank you for the change! it looks great. I left couple questions.

dbclient/WorkspaceClient.py
souravkhandelwal09: error_logger is per object type, and each Client deals with multiple objects. If we want to define it as global variables, we need to define multiple loggers (by action_type & object_type). Therefore I preferred getting the logger where needed and passing it around

dbclient/dbclient.py
souravkhandelwal09: If we want to add failures in dbclient, then we need to compliment all calls with action_type, object_type in order add failures to the right file. After auditing the code, it seemed like clients are printing custom failed logs from many places. In these cases, adding the error logging in dbclient will push duplicate error logs for the same failure and there is no easy way to avoid it.  Therefore I kept it as client responsibility.

I can add a util function which can be called to log error by action_type, object_type across all clients.

export_db.py
souravkhandelwal09: done

pipeline/pipeline.py
souravkhandelwal09: yeah verified that

kevinkim9264 (approved):

dbclient/dbclient.py
yubing: I mean we can log all errors into one file as the fallback just in case that we forget to add error logs for certain object types.

souravkhandelwal09: done

souravkhandelwal09:
> 

Done added

mrchristine (approved):
lgtm. I'd recommend that we update the README to include the example commands to run this pipeline. The description currently has a typo: 
```python migration-pipeline.py --profile e2-demo-src --export-pipeline --use-checkpoint --cluster-name API_Metastore_Work_Leave_Me_Alone``` 
The migration pipeline name uses underscore instead of dash.

PR #86, opened 2021-12-08
Make --cluster-name optional

kevinkim9264:
This PR makes the --cluster-name flag optional by making the downstream tasks launch cluster by itself.
Launch cluster is noop if the the cluster is already up and running.

Manual testing:
```
# Export (without --cluster-name):
python3 migration_pipeline.py --profile prod_src_miklos --export-pipeline --use-checkpoint


# Import (without --cluster-name):
python3 migration_pipeline.py --profile prod_dst_kevin --import-pipeline --use-checkpoint --session 20211207125516
```

### DIFF
--- a/data/aws_cluster.json
+++ b/data/aws_cluster.json
@@ -1,11 +1,10 @@
 {
   "num_workers": 1,
-  "cluster_name": "API_Metastore_Work_Leave_Me_Alone",
+  "cluster_name": "Workspace_Migration_Work_Leave_Me_Alone",
   "spark_version": "6.5.x-scala2.11",
   "aws_attributes": {
       "first_on_demand": 1,
       "availability": "SPOT_WITH_FALLBACK",
-      "zone_id": "us-west-2b",
       "spot_bid_price_percent": 100,
       "ebs_volume_count": 0
   },
--- a/dbclient/ClustersClient.py
+++ b/dbclient/ClustersClient.py
@@ -276,7 +276,9 @@ class ClustersClient(dbclient):
                 acl_args = {'access_control_list' : self.build_acl_args(data['access_control_list'])}
                 cid = self.get_cluster_id_by_name(cluster_name)
                 if cid is None:
-                    raise ValueError('Cluster id must exist in new env. Re-import cluster configs.')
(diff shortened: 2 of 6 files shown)


data/aws_cluster.json
```diff
@@ -1,11 +1,10 @@
 {
   "num_workers": 1,
-  "cluster_name": "API_Metastore_Work_Leave_Me_Alone",
+  "cluster_name": "Workspace_Migration_Work_Leave_Me_Alone",
```
kevinkim9264, line 3: Because this file is used whenever `launch_cluster()` is called (not just for Metastore work), we rename the cluster_name to be more generic

data/aws_cluster.json
```diff
@@ -1,11 +1,10 @@
...
   "spark_version": "6.5.x-scala2.11",
   "aws_attributes": {
       "first_on_demand": 1,
       "availability": "SPOT_WITH_FALLBACK",
-      "zone_id": "us-west-2b",
```
kevinkim9264, line 8: we should just not specify the zone_id so that this file can be used region agnostically.

yubing (approved):

kevinkim9264:
@mrchristine since Yubing "relies" on this PR change, we will  merge it. Please leave any comments if you have any concerns.

data/aws_cluster.json
mrchristine: We should use `auto` so it chooses a zone_id automatically in the future. By default Databricks will fix it to one AZ / zone_id. `auto` is a better option that we can improve upon later.

mrchristine (commented):
lgtm

data/aws_cluster.json
kevinkim9264: thank you! doesn’t databricks still randomly pick a zone_id? (or are you saying it’s hard-coded in terms of which zone id it will be used?) Hopefully, Databricks pick an available zone id though?

mrchristine: No, we've never randomly set the `zone_id`. We have a concept of `default`. If you go to any cluster UI, the defaults always set. If the API does not choose a zone, it will choose the default. The `auto` intelligently chooses the `zone_id` by looking at available ip space and if the instance isn't available, it can fallback to another zone.

PR #92, opened 2021-12-14
Added error handling for more clients

souravkhandelwal09:
**Description**
1. Updated error handling for other clients - ClustersClient, HiveClient, JobsClient, ScimClient, SecretsClient
2. Added logic to exclude error logging for expected errors - Resource_Already_Exists (when imports are done more than once)
3. Fixed the logic such that we can skip task before error handling check happens

**Testing**
Manually tested export and import

### DIFF
--- a/dbclient/ClustersClient.py
+++ b/dbclient/ClustersClient.py
@@ -2,7 +2,8 @@ import logging
 import os
 import re
 import time
-
+import logging_utils
+import wmconstants
 from dbclient import *
 
 
@@ -75,7 +76,7 @@ class ClustersClient(dbclient):
         import os
         real_path = os.path.dirname(os.path.realpath(__file__))
         if self.is_aws():
-            print("Updating cluster with: " + iam_role)
+            logging.info("Updating cluster with: " + iam_role)
             current_cluster_json = self.get(f'/clusters/get?cluster_id={cid}')
             run_properties = set(list(current_cluster_json.keys())) - self.create_configs
             for p in run_properties:
@@ -87,7 +88,7 @@ class ClustersClient(dbclient):
                 aws_conf = {'instance_profile_arn': iam_role}
             current_cluster_json['aws_attributes'] = aws_conf
             resp = self.post('/clusters/edit', current_cluster_json)
-            print(resp)
+            logging.info(resp)
             new_cid = self.wait_for_cluster(cid)
             return new_cid
         else:
@@ -128,7 +129,7 @@ class ClustersClient(dbclient):
             return clusters_list
 
     def get_execution_context(self, cid):
-        print("Creating remote Spark Session")
+        logging.info("Creating remote Spark Session")
         time.sleep(5)
         ec_payload = {"language": "python",
                       "clusterId": cid}
@@ -136,8 +137,8 @@ class ClustersClient(dbclient):
         # Grab the execution context ID
         ec_id = ec.get('id', None)
         if not ec_id:
-            print('Unable to establish remote session')
-            print(ec)
+            logging.info('Unable to establish remote session')
+            logging.info(ec)
             raise Exception("Remote session error")
         return ec_id
 
@@ -225,7 +226,7 @@ class ClustersClient(dbclient):
         cluster_log = self.get_export_dir() + log_file
         acl_cluster_log = self.get_export_dir() + acl_log_file
         if not os.path.exists(cluster_log):
-            print("No clusters to import.")
+            logging.info("No clusters to import.")
             return
         current_cluster_names = set([x.get('cluster_name', None) for x in self.get_cluster_list(False)])
         old_2_new_policy_ids = self.get_new_policy_id_dict()  # dict of {old_id : new_id}
@@ -235,7 +236,7 @@ class ClustersClient(dbclient):
                 cluster_conf = json.loads(line)
                 cluster_name = cluster_conf['cluster_name']
                 if cluster_name in current_cluster_names:
-                    print("Cluster already exists, skipping: {0}".format(cluster_name))
+                    logging.info("Cluster already exists, skipping: {0}".format(cluster_name))
                     continue
                 cluster_creator = cluster_conf.pop('creator_user_name')
                 if 'policy_id' in cluster_conf:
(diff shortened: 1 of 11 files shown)


pipeline/pipeline.py
```diff
@@ -71,11 +71,14 @@ def _run_task(self, task: AbstractTask):
...
-        failed_task_log = logging_utils.get_error_log_file(task.action_type, task.object_type, self._working_dir)
-        if os.path.exists(failed_task_log) and os.path.getsize(failed_task_log) > 0:
-            msg = f'{task.name} has failures. Refer to {failed_task_log} to see failures. Terminating pipeline.'
-            logging.info(msg)
-            raise RuntimeError(msg)
-        self._completed_steps.write(f'{task.name}')
+            end = timer()
+            logging.info(f'{task.name} Completed. Total time taken: {str(timedelta(seconds=end - start))}')
+            failed_task_log = logging_utils.get_error_log_file(task.action_type, task.object_type, self._working_dir)
+            if os.path.exists(failed_task_log) and os.path.getsize(failed_task_log) > 0:
```
yubing, line 77: If we want to use error logs to indicate errors (the errors that we need to fix always), how should we log warnings?

Previously, I thought we will treat all error logs as warnings. For errors, we can raise exceptions directly. My concern is treating all errors and warnings as hard errors will be too interruptive for users.

logging_utils.py
```diff
@@ -27,14 +28,19 @@ def get_error_logger(action_type, object_type, log_dir):
...
 def _get_log_dir(parent_dir):
     return parent_dir + "/app_logs"
 
-def log_reponse_error(error_logger, response, error_msg=None):
-    if 'error_code' in response or 'error' in response:
+
+def log_reponse_error(error_logger, response, error_msg=None, error_ignore_list=[]):
+    if ('error_code' in response and response['error_code'] not in error_ignore_list) \
+            | ('error' in response) \
+            | (response.get('resultType', None) == 'error' and 'already exists' not in response.get('summary', None)):
```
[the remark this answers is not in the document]
kevinkim9264: already exists can happen when the source workspace is already used and has the same resource, or the import script is run again which is very plausible. I think this should be skipped, but should be a **warning**.

pipeline/pipeline.py
kevinkim9264: This is a bit different from treating all errors and warnings as hard errors, because we wait for the entire task to finish. For the hard errors, it would just raise exception and stop in the middle of the task phase. For this one, we still let the task finish but before moving onto the next task, we interrupt the pipeline and ask users to look at the error log files, so this is almost the warning files. 

But I agree that this change can feel like a regression from the user experience perspective. We would like the invokers to make a discerning decisions, but because the invokers will be non-expert at times, we might want to do our best to differentiate between hard errors vs warnings and not let the pipeline fail if these are just warnings.

We might want to discuss on this together @yubing  @souravkhandelwal09

dbclient/SecretsClient.py
```diff
@@ -132,37 +139,28 @@ def import_all_secrets(self, log_dir='secret_scopes/'):
                     create_scope_args['initial_manage_principal'] = 'users'
                 other_permissions = self.get_all_other_permissions(scope_name, scopes_acl_dict)
                 create_resp = self.post('/secrets/scopes/create', create_scope_args)
-                print(create_resp)
+                logging_utils.log_reponse_error(error_logger, create_resp, None, ['RESOURCE_ALREADY_EXISTS'])
```
kevinkim9264, line 142: can we put `error_ignore_list=['RESOURCE...']`?

logging_utils.py
```diff
@@ -27,14 +28,19 @@ def get_error_logger(action_type, object_type, log_dir):
...
     return f"{_get_log_dir(parent_dir)}/failed_{action_type}_{object_type}.log"
 
+
 def _get_log_dir(parent_dir):
     return parent_dir + "/app_logs"
 
-def log_reponse_error(error_logger, response, error_msg=None):
-    if 'error_code' in response or 'error' in response:
+
+def log_reponse_error(error_logger, response, error_msg=None, error_ignore_list=[]):
```
kevinkim9264, line 40: If we are intending to ignore the ALREADY_EXIST error in the below logic, should we just put it into the default value of the error_ignore_list? (['RESOURCE_ALREADY_EXISTS'])

logging_utils.py
```diff
@@ -27,14 +28,19 @@ def get_error_logger(action_type, object_type, log_dir):
...
+    if ('error_code' in response and response['error_code'] not in error_ignore_list) \
+            | ('error' in response) \
```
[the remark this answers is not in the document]
souravkhandelwal09: `or` and `|` are the same. I changed it to `or` to be consistent.
Added ` response['error'] in error_ignore_list`

logging_utils.py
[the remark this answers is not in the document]
souravkhandelwal09: added a catch all error in dbclient and logging it as WARN

pipeline/pipeline.py
souravkhandelwal09: Had a discussion with @yubing , I added all a catch all warn for errors in dbclient.

dbclient/SecretsClient.py
souravkhandelwal09: ✔️

logging_utils.py
souravkhandelwal09: done

mrchristine (approved):
lgtm!

wmconstants.py
```diff
@@ -12,7 +12,7 @@
 CLUSTER_OBJECT = "clusters"
 INSTANCE_POOL_OBJECT = "instance_pools"
 JOB_OBJECT = "jobs"
-SECTEC_OBJECT = "secrets"
```
kevinkim9264, line 15: √

kevinkim9264 (approved):

dbclient/ClustersClient.py
```diff
@@ -372,7 +372,7 @@ def import_cluster_policies(self, log_file='cluster_policies.log', acl_log_file=
                                    'definition': policy_conf['definition']}
                     resp = self.post('/policies/clusters/create', create_args)
                     ignore_error_list = ['INVALID_PARAMETER_VALUE']
-                    logging_utils.log_reponse_error(error_logger, resp, None, ignore_error_list)
+                    logging_utils.log_reponse_error(error_logger, resp, error_msg=None, ignore_error_list=ignore_error_list)
```
yubing, line 375: Nit: error_msg is not necessary if the value is equal to default value.

yubing (approved):
LGTM with nit.

PR #99, opened 2022-01-26
[PLAT-46718] Export ml experiment

kevinkim9264:
usage:
`python3 export_db.py --profile $PROFILE --mlflow_experiments`

Tests:
```
python3 export_db.py --profile dogfood --mlflow_experiments
Exporting MLflow experiments.
2022-01-25,16:53:43;INFO;Complete MLflow Experiments Export Time: 0:00:06.048322

// writes logs/mlflow_experiments.log file (with 3954 items)
{"artifact_location": "dbfs:/databricks/mlflow/10065689", "experiment_id": "10065689", "lifecycle_stage": "active", "name": "/Users/aaron@databricks.com/Movie Recommendations", "tags": {"mlflow.ownerId": "", "mlflow.experimentType": "NOTEBOOK"}}
{"artifact_location": "dbfs:/databricks/mlflow/12174144", "experiment_id": "12174144", "lifecycle_stage": "active", "name": "/Users/aaron@databricks.com/Basic MLflow", "tags": {"mlflow.ownerId": "", "mlflow.experimentType": "NOTEBOOK"}}
{"artifact_location": "dbfs:/databricks/mlflow/12557050", "experiment_id": "12557050", "lifecycle_stage": "active", "name": "/Users/aaron@databricks.com/A", "tags": {"mlflow.ownerId": "", "mlflow.experimentType": "NOTEBOOK"}}
{"artifact_location": "dbfs:/databricks/mlflow/13430347", "experiment_id": "13430347", "lifecycle_stage": "active", "name": "/Shared/experiments/Default Experiment", "tags": {"mlflow.note.content": "test123 coool", "mlflow.ownerId": "101003", "mlflow.experimentType": "MLFLOW_EXPERIMENT", "mlflow.ownerEmail": "mlflow_migrator+dbadmin@databricks.com"}}
......
```

I will not add ML object exports to migration task yet, and leave the invokers to opt into migrate ml objects (since most of the customers do not migrate ml objects)

### DIFF
--- a/dbclient/MLFlowClient.py
+++ b/dbclient/MLFlowClient.py
@@ -26,8 +26,6 @@ class MLFlowClient:
         experiments_logfile = mlflow_experiments_dir + log_file
         with open(experiments_logfile, 'w') as fp:
             for experiment in experiments:
-                # TODO(kevin): do try catch on get_experiment call
-                experiment_metadata = self.client.get_experiment(experiment.experiment_id)
-                fp.write(json.dumps(dict(experiment_metadata)) + '\n')
+               fp.write(json.dumps(dict(experiment)) + '\n')
         end = timer()
         logging.info("Complete MLflow Experiments Export Time: " + str(timedelta(seconds=end - start)))
--- a/dbclient/MLFlowClient.py
+++ b/dbclient/MLFlowClient.py
@@ -22,6 +22,10 @@ class MLFlowClient:
         )
         os.makedirs(mlflow_experiments_dir, exist_ok=True)
         start = timer()
+        # We do not do any pagination since ST workspaces do not have that many experiments count.
+        # Max is ~6k experiments for the moment.
+        # Consider using pagination(https://www.mlflow.org/docs/latest/python_api/mlflow.tracking.html) if
+        # a workspace has explosive number of experiments. (e.g. 200K)
         experiments = self.client.list_experiments(view_type=ViewType.ALL)
         experiments_logfile = mlflow_experiments_dir + log_file
         with open(experiments_logfile, 'w') as fp:
--- a/dbclient/MLFlowClient.py
+++ b/dbclient/MLFlowClient.py
@@ -1,0 +1,33 @@
+import os
+import json
+from datetime import timedelta
+from timeit import default_timer as timer
+import logging
+import logging_utils
+import mlflow
+from mlflow.tracking import MlflowClient
+from mlflow.entities import ViewType
+import wmconstants
+
+class MLFlowClient:
+    def __init__(self, configs, checkpoint_service):
+        self._checkpoint_service = checkpoint_service
+        self.export_dir = configs['export_dir']
+        self.client = MlflowClient(f"databricks://{configs['profile']}")
+
+    def export_mlflow_experiments(self, log_file='mlflow_experiments.log', log_dir=None):
+        mlflow_experiments_dir = log_dir if log_dir else self.export_dir
+        error_logger = logging_utils.get_error_logger(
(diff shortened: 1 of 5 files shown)


dbclient/MLFlowClient.py
```diff
@@ -0,0 +1,31 @@
...
+        self.client = MlflowClient(f"databricks://{configs['profile']}")
+
+    def export_mlflow_experiments(self, log_file='mlflow_experiments.log', log_dir=None):
+        mlflow_experiments_dir = log_dir if log_dir else self.export_dir
+        error_logger = logging_utils.get_error_logger(
+            wmconstants.WM_EXPORT, wmconstants.MLFLOW_EXPERIMENT_OBJECT, self.export_dir
+        )
+        os.makedirs(mlflow_experiments_dir, exist_ok=True)
+        start = timer()
+        experiments = self.client.list_experiments(view_type=ViewType.ALL)
```
souravkhandelwal09, line 25: what is the count of ml experiments for the largest customer ? If the count is large, pagination would be better. Do they have a paginated api we can use ?

kevinkim9264: ah yeah, I didn't leave the comments, but we don't have to worry about as the count is relatively small. 
(the biggest is from shard-dogfood which only has ~6k)
https://logfood.cloud.databricks.com/#notebook/11746207/command/12643484

Let me leave the comment, but Sid can confirm

souravkhandelwal09 (approved):

dbclient/MLFlowClient.py
```diff
@@ -0,0 +1,35 @@
...
+        start = timer()
+        # We do not do any pagination since ST workspaces do not have that many experiments count.
+        # Max is ~6k experiments for the moment.
+        # Consider using pagination(https://www.mlflow.org/docs/latest/python_api/mlflow.tracking.html) if
+        # a workspace has explosive number of experiments. (e.g. 200K)
+        experiments = self.client.list_experiments(view_type=ViewType.ALL)
+        experiments_logfile = mlflow_experiments_dir + log_file
+        with open(experiments_logfile, 'w') as fp:
+            for experiment in experiments:
+               fp.write(json.dumps(dict(experiment)) + '\n')
```
smurching, line 33: I'd suggest writing the experiments in `pickle` format to make it easier to read them back + access their fields during import, but json is fine too!

dbclient/MLFlowClient.py
smurching: Yeah, no need to paginate for ST!

smurching: We may need to add pagination later for large MT workspaces etc, but for this initial ST -> E2 migration this seems fine.

smurching (approved):
LGTM with just a suggestion on the persistence format for the MLflow experiments

dbclient/MLFlowClient.py
kevinkim9264: @smurching is it easier to read because I can access the info as fields of a python object? (instead of reading json?)
It seems like I have to call `client.create_experiment` and `set_experiment_tag` anyway and hence have to access the fields separately (which we can do with json too anyway)

Because the json has clear texts which is useful for simple debugging, unit testing, etc and also because pickle binary could actually have an offending code which can be accidentally executed while deserializing, json seems safer unless we have a strong reason to use pickle. (e.g. we can directly reuse the deserialized experiment object to import the data)

smurching: Yep, mainly thinking it's easier to access the Python object fields! Thanks for the catch RE security, JSON seems fine

PR #103, opened 2022-02-03
Parallelize apis for workspace export

kevinkim9264:
This PR parallelizes APIs for workspace exports. Specifically,

```
python3 export_db.py --profile $PROFILE --workspace --num-parallel 20
python3 export_db.py --profile $PROFILE --download --num-parallel 20
```

Some benchmarks:
```
python3 export_db.py --profile shardqa --workspace --num-parallel $Num
current:         7 minutes
5 threads:     6 minutes
10 threads:   6 minutes
15 threads:   6 minutes
20 threads:  5 minutes
```

```
python3 export_db.py --profile shardqa --download --num-parallel $Num
current:  46 minutes
5 threads:      8 minutes
10 threads:    6 minutes
15 threads:    5 minutes
20 threads:   3 minutes
```

Observations:
--workspace does not benefit too much from the parallelism because there is only one thread writing to a file (to prevent race condition in file writing). 
Note: threads are not blocked by each other as they all independently get the results from APIs and put into the SafeWriter queue. It's just that it takes a long time for the single thread of SafeWriter to write all the contents to a file. (so it would still benefit some amount, because API calls are parallelized)

--download benefits alot from the number of parallelism because each of the threads writes to an independent file and hence not bottlenecked by file writings.

tldr is that in the current design, the parallelism benefits kick in only when we output to multiple files. If we really want to improve the parallelism benefits for single file writing cases, we have to change the script logics to deal with multiple files instead of single files. (e.g. instead of user_workspace.log    user_workspace_1.log, user_workspace_2.log, user_workspace_3.log, user_workspace_4.log, etc)


mrchristine:
lgtm and tests show performance improvements for parallelized operations.

dbclient/WorkspaceClient.py
```diff
@@ -260,34 +271,35 @@ def download_notebooks(self, ws_log_file='user_workspace.log', ws_dir='artifacts
...
             # notebook log metadata file now contains object_id to help w/ ACL exports
             # pull the path from the data to download the individual notebook contents
-            for notebook_data in fp:
-                notebook_path = json.loads(notebook_data).get('path', None).rstrip('\n')
-                dl_resp = self.download_notebook_helper(notebook_path, checkpoint_notebook_set, notebook_error_logger,
-                                                        export_dir=self.get_export_dir() + ws_dir)
-                if 'error' not in dl_resp:
-                    num_notebooks += 1
+            with ThreadPoolExecutor(max_workers=num_parallel) as executor:
+                futures = {executor.submit(self.download_notebook_helper, notebook_data, checkpoint_notebook_set, notebook_error_logger, self.get_export_dir() + ws_dir): notebook_data for notebook_data in fp}
```
souravkhandelwal09, line 275: more python like would be 
`[executor.submit(self.download_notebook_helper, notebook_data, checkpoint_notebook_set, notebook_error_logger, self.get_export_dir() + ws_dir) for notebook_data in fp]`

dbclient/WorkspaceClient.py
```diff
@@ -388,22 +400,31 @@ def log_all_workspace_items(self, ws_path, workspace_log_writer, libs_log_writer
                 libs_log_writer.write(json.dumps(y) + '\n')
```
souravkhandelwal09, line 400: We can also parallelize this ?

souravkhandelwal09 (commented):
We also need to make the checkpoint files ThreadSafeWriter

dbclient/WorkspaceClient.py
```diff
@@ -176,9 +178,18 @@ def export_user_home(self, username, local_export_dir):
         if num_of_nbs != num_of_nbs_dl:
             print(f"Notebooks logged != downloaded. Check the failed download file at: {user_export_dir}")
         print(f"Exporting the notebook permissions for {username}")
-        self.log_acl_to_file('notebooks', 'user_workspace.log', 'acl_notebooks.log', 'failed_acl_notebooks.log')
+        acl_notebooks_writer = ThreadSafeWriter("acl_notebooks.log", "w")
+        try:
+            self.log_acl_to_file('notebooks', 'user_workspace.log', acl_notebooks_writer, 'failed_acl_notebooks.log')
```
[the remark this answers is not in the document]
kevinkim9264: we just have to do the error_logger since python logging libraries are already thread-safe: https://docs.python.org/2/library/logging.html#thread-safety

dbclient/WorkspaceClient.py
```diff
@@ -176,9 +178,18 @@ def export_user_home(self, username, local_export_dir):
...
+        finally:
+            acl_notebooks_writer.close()
+
         print(f"Exporting the directories permissions for {username}")
-        self.log_acl_to_file('directories', 'user_dirs.log', 'acl_directories.log', 'failed_acl_directories.log')
+        acl_directories_writer = ThreadSafeWriter("acl_directories.log", "w")
+        try:
+            self.log_acl_to_file('directories', 'user_dirs.log', acl_directories_writer, 'failed_acl_directories.log')
```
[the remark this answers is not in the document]
kevinkim9264: ditto

dbclient/WorkspaceClient.py
kevinkim9264: No need to because they are only doing the writing part to the same file. (there is 0 performance improvement when it comes to writing to same files from multiple threads)

dbclient/WorkspaceClient.py
kevinkim9264: √

PR #106, opened 2022-02-10
Support mlflow_experiment import

kevinkim9264:
Supports mlflow_experiment import

```
python3 import_db.py --profile mwc_dst --mlflow-experiments --use-checkpoint
```

The command consumes `mlflow_experiments.log` (the output of export_db.py --mlflow-experiments) and creates the experiment objects in the destination workspace. Also it outputs `mlflow_experiments_id_map.log` which has a format of
```
{"old_id": "3825356326110958", "new_id": "2025369099255681"}
{"old_id": "3825356326110957", "new_id": "2025369099255680"}
```
which will be consumed by run import later.

This PR also makes the checkpointer use ThreadSafeWriter

Notes:
- mlflow_experiments_id_map.log only appends instead of rewriting. This is to ensure that upon failures and retries, it still keeps the already imported experiments.
- meaning, one should never remove the mlflow_experiments_id_map.log after running the import command in order to not lose the already imported experiment objects' id mapping.

### DIFF
--- a/dbclient/MLFlowClient.py
+++ b/dbclient/MLFlowClient.py
@@ -7,7 +7,9 @@ import logging_utils
 import mlflow
 from mlflow.tracking import MlflowClient
 from mlflow.entities import ViewType
+from mlflow.exceptions import RestException
 import wmconstants
+from thread_safe_writer import ThreadSafeWriter
 
 class MLFlowClient:
     def __init__(self, configs, checkpoint_service):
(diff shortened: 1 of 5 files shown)


dbclient/MLFlowClient.py
```diff
@@ -33,3 +37,62 @@ def export_mlflow_experiments(self, log_file='mlflow_experiments.log', log_dir=N
...
+            checkpointer.write(id)
+
+    def _cleanse_artifact_location(self, artifact_location):
+        """
+        There are some paths that are not allowed to be artifact_location. In those cases, we should use None as the
+        artifact_location when creating experiment objects.
+        """
+        if artifact_location == None or \
+                artifact_location.startswith("dbfs:/databricks/mlflow-tracking") or \
+                artifact_location.startswith("dbfs:/databricks/mlflow"):
```
smurching, line 96:
```suggestion
                artifact_location.startswith("dbfs:/databricks/mlflow/"):
```

Nit: just to capture the full prefix in this case

dbclient/MLFlowClient.py
```diff
@@ -33,3 +37,62 @@ def export_mlflow_experiments(self, log_file='mlflow_experiments.log', log_dir=N
...
+            # checkpoint the original id
+            checkpointer.write(id)
+
+    def _cleanse_artifact_location(self, artifact_location):
+        """
+        There are some paths that are not allowed to be artifact_location. In those cases, we should use None as the
+        artifact_location when creating experiment objects.
+        """
+        if artifact_location == None or \
+                artifact_location.startswith("dbfs:/databricks/mlflow-tracking") or \
```
smurching, line 95:
```suggestion
                artifact_location.startswith("dbfs:/databricks/mlflow-tracking/") or \
```

smurching (commented):
Mostly LGTM! Just minor comments

Issue #119, opened 2022-02-24
export a single users workspace fails with exception


paulg-mscience:
When I run a command similar to this:
```
python export_db.py --profile DEMO --export-home example@foobar.com
```
I get the following exception:
```
Traceback (most recent call last):
  File "export_db.py", line 308, in <module>
    main()
  File "export_db.py", line 262, in main
    ws_c.export_user_home(username, 'user_exports')
  File "/home/ec2-user/github/databricks-migrate/dbclient/WorkspaceClient.py", line 181, in export_user_home
    workspace_log_writer = ThreadSafeWriter(self.get_export_dir() + 'user_workspace.log', "a")
  File "/home/ec2-user/github/databricks-migrate/thread_safe_writer.py", line 17, in __init__
    self.filewriter = open(*args)
FileNotFoundError: [Errno 2] No such file or directory: 'logs/user_exports/example@foobar.com/user_workspace.log'
```

mrchristine:
@kevinkim9264 @souravkhandelwal09 can we take a look at this?

kevinkim9264:
@paulg-mscience you can try this https://github.com/databrickslabs/migrate/pull/120

paulg-mscience:
@kevinkim9264 You fixed the first exception. Thank you. Now I am encountering a new exception.

**Start with empty or non-existing logs folder.**

```
python export_db.py --profile DEMO --export-home example@foobar.com
```

```
Traceback (most recent call last):
  File "export_db.py", line 308, in <module>
    main()
  File "export_db.py", line 262, in main
    ws_c.export_user_home(username, 'user_exports')
  File "/home/ec2-user/github/databricks-migrate/dbclient/WorkspaceClient.py", line 195, in export_user_home
    num_of_nbs_dl = self.download_notebooks(ws_dir='user_artifacts/')
  File "/home/ec2-user/github/databricks-migrate/dbclient/WorkspaceClient.py", line 292, in download_notebooks
    notebook_error_logger = logging_utils.get_error_logger(
  File "/home/ec2-user/github/databricks-migrate/logging_utils.py", line 25, in get_error_logger
    error_handler = logging.FileHandler(failed_log_file, 'w+')
  File "/home/ec2-user/.pyenv/versions/3.8.7/lib/python3.8/logging/__init__.py", line 1143, in __init__
    StreamHandler.__init__(self, self._open())
[...]

PR #123, opened 2022-02-25
Fix2 of export-user-home

kevinkim9264:
https://github.com/databrickslabs/migrate/issues/119

`python3 export_db.py --profile shardqa --export-home kevinkim@databricks.com` works

Also tested for no regression for pipeline workflow:
```
date; time python3 migration_pipeline.py --profile XXX --export-pipeline --use-checkpoint --num-parallel 20 --skip-tasks 'metastore' 'metastore_table_acls'
```

### DIFF
--- a/logging_utils.py
+++ b/logging_utils.py
@@ -22,6 +22,8 @@ def get_error_logger(action_type, object_type, log_dir):
     logger = logging.getLogger("workspace_migration")
 
     failed_log_file = get_error_log_file(action_type, object_type, log_dir)
+    os.makedirs(_get_log_dir(log_dir), exist_ok=True)
+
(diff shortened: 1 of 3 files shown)


logging_utils.py
```diff
@@ -22,6 +22,8 @@ def get_error_logger(action_type, object_type, log_dir):
     logger = logging.getLogger("workspace_migration")
 
     failed_log_file = get_error_log_file(action_type, object_type, log_dir)
+    os.makedirs(_get_log_dir(log_dir), exist_ok=True)
```
kevinkim9264, line 25: This is the real fix to the reported issue

paulg-mscience: Yes, this fixes the issue. Thank you!

Issue #125, opened 2022-03-01
Add checkpointing to job import


paulg-mscience:
We have 1250 jobs. It takes approx 1 hour to import these jobs. It would be nice to have checkpoint for job import. Currently if job import fails, re-running the import results in duplicate jobs being created.

kevinkim9264:
yeah, we can technically do this using the following way:
1. Import job and if successful checkpoint by the job_id
2. On the rerun, if the exported job_id is already in the checkpoint, skip

We can do the same thing for cluster import with cluster_id.
(We can't use job name or cluster name since the names are not unique)

I am really backloaded with other works, but I will try to get to it some time later

kevinkim9264:
@paulg-mscience https://github.com/databrickslabs/migrate/pull/131 is the work

paulg-mscience:
Implemented here ... https://github.com/databrickslabs/migrate/commit/3adc23d539a8d61131faed5001a10cb86c26bdc2

PR #131, opened 2022-03-10
[Issue 125][PLAT-50595] Add checkpointer for cluster and jobs import

kevinkim9264:
This PR adds checkpointer functionality to cluster and jobs import as raised here: https://github.com/databrickslabs/migrate/issues/125

With this PR, if clusters/jobs import fail or interrupted in the middle, the retry will skip the already imported objects.

Test:
```
date; time python3 migration_pipeline.py --profile mwc_dst --import-pipeline --use-checkpoint --num-parallel 10 --skip-tasks 'metastore' 'metastore_table_acls' 'workspace_acls' --session M20220309125843
```
After the import is done, I manually removed "import_cluster" and "import_jobs" from `logs/M20220309125843/checkpoint/import_tasks.log` so that the rerun of the pipeline executes the jobs and clusters tasks. But because each of the objects are checkpointed, the task will skip the already imported ones

e.g. another run of the pipeline would show:
cluster import:
```
2022-03-09,14:12:21;INFO;/cluster-policies/9A5F25BA4700004A found in checkpoint
2022-03-09,14:12:21;INFO;/cluster-policies/9A5F25BA47000049 found in checkpoint
2022-03-09,14:12:21;INFO;/cluster-policies/9A5F25BA4700000A found in checkpoint
2022-03-09,14:12:21;INFO;/cluster-policies/9A5F25BA47000009 found in checkpoint
2022-03-09,14:12:21;INFO;/cluster-policies/9A5F25BA47000008 found in checkpoint
Get: https://xxxxxxxxxxxxxxxx/api/2.0/clusters/list
Get: https://xxxxxxxxxxxxxxxx/api/2.0/policies/clusters/list
2022-03-09,14:12:22;INFO;Cluster already exists, skipping: Workspace_Migration_Work_Leave_Me_Alone
2022-03-09,14:12:22;INFO;0225-034621-fyv7t2c found in checkpoint
2022-03-09,14:12:22;INFO;Cluster already exists, skipping: pools_cluster
2022-03-09,14:12:22;INFO;Cluster already exists, skipping: Build Tests
2022-03-09,14:12:22;INFO;Cluster already exists, skipping: cluster_policy_tests
2022-03-09,14:12:22;INFO;0211-174010-i88s105s found in checkpoint
2022-03-09,14:12:22;INFO;Cluster already exists, skipping: API_Table_ACL_Work_Leave_Me_Alone
Get: https://xxxxxxxxxxxxxxxx/api/2.0/clusters/list
The old cluster id 1207-190206-21uier75 with the original_creator of kevinkim+dbadmin@databricks.com does not get logged for EditClusterOwner due to some problems.
2022-03-09,14:12:22;INFO;/clusters/1207-190206-21uier75 found in checkpoint
2022-03-09,14:12:22;INFO;/clusters/0225-034621-fyv7t2c found in checkpoint
2022-03-09,14:12:22;INFO;/clusters/0225-232459-sited260 found in checkpoint
2022-03-09,14:12:22;INFO;/clusters/0225-232454-roe258 found in checkpoint
2022-03-09,14:12:22;INFO;/clusters/0225-232456-vowel259 found in checkpoint
2022-03-09,14:12:22;INFO;/clusters/0211-174010-i88s105s found in checkpoint
2022-03-09,14:12:22;INFO;/clusters/0127-215202-ork49kx9 found in checkpoint
2022-03-09,14:12:22;INFO;import_clusters Completed. Total time taken: 0:00:01.374465
```

jobs import:
```
2022-03-09,16:11:42;INFO;Start import_jobs
Get: https://xxxxxxxxxxxxxxxx/api/2.0/clusters/list
Get: https://xxxxxxxxxxxxxxxx/api/2.0/policies/clusters/list
2022-03-09,16:11:43;INFO;1 found in checkpoint
2022-03-09,16:11:43;INFO;2 found in checkpoint
2022-03-09,16:11:43;INFO;4 found in checkpoint
2022-03-09,16:11:43;INFO;7 found in checkpoint
2022-03-09,16:11:43;INFO;8 found in checkpoint
2022-03-09,16:11:43;INFO;9 found in checkpoint
2022-03-09,16:11:43;INFO;1099636976528903 found in checkpoint
Get: https://xxxxxxxxxxxxxxxx.cloud.databricks.com/api/2.0/jobs/list
Get: https://xxxxxxxxxxxxxxxx.cloud.databricks.com/api/2.1/jobs/list?expand_tasks=true&offset=0&limit=25
Get: https://xxxxxxxxxxxxxxxx.cloud.databricks.com/api/2.1/jobs/list?expand_tasks=true&offset=25&limit=25
Get: https://xxxxxxxxxxxxxxxx.cloud.databricks.com/api/2.1/jobs/list?expand_tasks=true&offset=50&limit=25
Get: https://xxxxxxxxxxxxxxxx.cloud.databricks.com/api/2.1/jobs/list?expand_tasks=true&offset=75&limit=25
Get: https://xxxxxxxxxxxxxxxx.cloud.databricks.com/api/2.1/jobs/list?expand_tasks=true&offset=100&limit=25
Get: https://xxxxxxxxxxxxxxxx.cloud.databricks.com/api/2.1/jobs/list?expand_tasks=true&offset=125&limit=25
2022-03-09,16:11:45;INFO;/jobs/1 found in checkpoint
2022-03-09,16:11:45;INFO;/jobs/2 found in checkpoint
2022-03-09,16:11:45;INFO;/jobs/4 found in checkpoint
2022-03-09,16:11:45;INFO;/jobs/7 found in checkpoint
2022-03-09,16:11:45;INFO;/jobs/8 found in checkpoint
2022-03-09,16:11:45;INFO;/jobs/9 found in checkpoint
2022-03-09,16:11:45;INFO;/jobs/1099636976528903 found in checkpoint
Get: https://xxxxxxxxxxxxxxxx.cloud.databricks.com/api/2.0/jobs/list
Get: https://xxxxxxxxxxxxxxxx.cloud.databricks.com/api/2.1/jobs/list?expand_tasks=true&offset=0&limit=25
Get: https://xxxxxxxxxxxxxxxx.cloud.databricks.com/api/2.1/jobs/list?expand_tasks=true&offset=25&limit=25
Get: https://xxxxxxxxxxxxxxxx.cloud.databricks.com/api/2.1/jobs/list?expand_tasks=true&offset=50&limit=25
Get: https://xxxxxxxxxxxxxxxx.cloud.databricks.com/api/2.1/jobs/list?expand_tasks=true&offset=75&limit=25
Get: https://xxxxxxxxxxxxxxxx.cloud.databricks.com/api/2.1/jobs/list?expand_tasks=true&offset=100&limit=25
Get: https://xxxxxxxxxxxxxxxx.cloud.databricks.com/api/2.1/jobs/list?expand_tasks=true&offset=125&limit=25
2022-03-09,16:11:47;INFO;normal job found in checkpoint
```

Also tested using single task runs:
```
python3 import_db.py --profile mwc_dst --jobs --use-checkpoint --num-parallel 20
python3 import_db.py --profile mwc_dst --clusters --use-checkpoint --num-parallel 20
```


dbclient/JobsClient.py
```diff
@@ -143,6 +148,8 @@ def adjust_ids_for_cluster(settings): #job_settings or task_settings
         with open(jobs_log, 'r') as fp:
             for line in fp:
                 job_conf = json.loads(line)
+                if 'job_id' in job_conf and checkpoint_job_configs_set.contains(str(job_conf['job_id'])):
```
kevinkim9264, line 151: need to do this `str(...)`  because otherwise the job_id is recognized as integer which becomes str vs int which never matches. (in which case, the checkpoint never recognizes that the job_id is already checkpointed)

gregwood-db: @kevinkim9264 I noticed something similar happened with object/directory IDs when doing comparisons in my pruning script... is there anywhere else that this might happen?

paulg-mscience:
@kevinkim9264 Thanks for this PR. I will test it soon!

dbclient/JobsClient.py
kevinkim9264: @gregwood-db can you explain the "something similar" happened part?
I just checked the checkpointed files for "import_notebooks.log" file and and they are all String. 
This would happen if the id we checkpoint is pure integer

kevinkim9264:
> @kevinkim9264 Thanks for this PR. I will test it soon!

thanks. let me know how it goes!

migration_pipeline.py
```diff
@@ -79,16 +79,16 @@ def build_export_pipeline(client_config, checkpoint_service, args) -> Pipeline:
...
-    export_jobs = pipeline.add_task(JobsExportTask(client_config, args, wmconstants.JOBS in skip_tasks), [export_instance_pools])
+    export_secrets = pipeline.add_task(SecretExportTask(client_config, args, checkpoint_service, wmconstants.SECRETS in skip_tasks), [export_groups])
+    export_clusters = pipeline.add_task(ClustersExportTask(client_config, args, checkpoint_service, wmconstants.CLUSTERS in skip_tasks), [export_secrets])
+    export_instance_pools = pipeline.add_task(InstancePoolsExportTask(client_config, args, checkpoint_service, wmconstants.INSTANCE_POOLS in skip_tasks), [export_clusters])
+    export_jobs = pipeline.add_task(JobsExportTask(client_config, args, checkpoint_service, wmconstants.JOBS in skip_tasks), [export_instance_pools])
     export_metastore = pipeline.add_task(MetastoreExportTask(client_config, checkpoint_service, args, wmconstants.METASTORE in skip_tasks), [export_groups])
     export_metastore_table_acls = pipeline.add_task(MetastoreTableACLExportTask(client_config, args, wmconstants.METASTORE_TABLE_ACLS in skip_tasks), [export_metastore])
```
paulg-mscience, line 93: Need checkpoint_service param here.

migration_pipeline.py
```diff
@@ -118,15 +118,15 @@ def build_import_pipeline(client_config, checkpoint_service, args) -> Pipeline:
...
-    import_jobs = pipeline.add_task(JobsImportTask(client_config, args, wmconstants.JOBS in skip_tasks), [import_instance_pools])
+    import_secrets = pipeline.add_task(SecretImportTask(client_config, checkpoint_service, wmconstants.SECRETS in skip_tasks), [import_groups])
+    import_clusters = pipeline.add_task(ClustersImportTask(client_config, args, checkpoint_service, wmconstants.CLUSTERS in skip_tasks), [import_secrets])
+    import_instance_pools = pipeline.add_task(InstancePoolsImportTask(client_config, args, checkpoint_service, wmconstants.INSTANCE_POOLS in skip_tasks), [import_clusters])
+    import_jobs = pipeline.add_task(JobsImportTask(client_config, args, checkpoint_service, wmconstants.JOBS in skip_tasks), [import_instance_pools])
     import_metastore = pipeline.add_task(MetastoreImportTask(client_config, checkpoint_service, args, wmconstants.METASTORE in skip_tasks), [import_groups])
     import_metastore_table_acls = pipeline.add_task(MetastoreTableACLImportTask(client_config, args, wmconstants.METASTORE_TABLE_ACLS in skip_tasks), [import_metastore])
```
paulg-mscience, line 131: Need `checkpoint_service` param here for MetastoreTableACLImportTask

migration_pipeline.py
kevinkim9264: Ah Thank you!

kevinkim9264:
@paulg-mscience thanks for the review, could you test again?

migration_pipeline.py
```diff
@@ -79,18 +79,18 @@ def build_export_pipeline(client_config, checkpoint_service, args) -> Pipeline:
...
-    export_metastore = pipeline.add_task(MetastoreExportTask(client_config, checkpoint_service, args, wmconstants.METASTORE in skip_tasks), [export_groups])
-    export_metastore_table_acls = pipeline.add_task(MetastoreTableACLExportTask(client_config, args, wmconstants.METASTORE_TABLE_ACLS in skip_tasks), [export_metastore])
+    export_secrets = pipeline.add_task(SecretExportTask(client_config, args, checkpoint_service, wmconstants.SECRETS in skip_tasks), [export_groups])
+    export_clusters = pipeline.add_task(ClustersExportTask(client_config, args, checkpoint_service, wmconstants.CLUSTERS in skip_tasks), [export_secrets])
+    export_instance_pools = pipeline.add_task(InstancePoolsExportTask(client_config, args, checkpoint_service, wmconstants.INSTANCE_POOLS in skip_tasks), [export_clusters])
+    export_jobs = pipeline.add_task(JobsExportTask(client_config, args, checkpoint_service, wmconstants.JOBS in skip_tasks), [export_instance_pools])
+    export_metastore = pipeline.add_task(MetastoreExportTask(client_config, checkpoint_service, args, checkpoint_service, wmconstants.METASTORE in skip_tasks), [export_groups])
```
paulg-mscience, line 92: Oops. Now MetastoreExportTask has two `checkpoint_service` params.

kevinkim9264: ah.. I hate python :( Just fixed it

paulg-mscience:
@kevinkim9264 I encountered an error in my testing.

I executed this command:
```
date; time python migration_pipeline.py --import-pipeline --profile scratch-green --session M20220310200413 --use-checkpoint --debug --num-parallel 10 --skip-tasks 'metastore' 'metastore_table_acls' 'workspace_acls' 'notebooks' 2>&1 | tee -a tmp.txt
```

The above command failed with http 429 as I had predicted. 
```
post: https://mscience-scratch-green.cloud.databricks.com/api/2.0/jobs/update
Traceback (most recent call last):
  File "migration_pipeline.py", line 328, in <module>
    main()
  File "migration_pipeline.py", line 324, in main
    pipeline.run()
  File "/home/ec2-user/github/databricks-migrate/pipeline/pipeline.py", line 64, in run
    future.result()
  File "/home/ec2-user/.pyenv/versions/3.8.7/lib/python3.8/concurrent/futures/_base.py", line 439, in result
    return self.__get_result()
  File "/home/ec2-user/.pyenv/versions/3.8.7/lib/python3.8/concurrent/futures/_base.py", line 388, in __get_result
    raise self._exception
  File "/home/ec2-user/.pyenv/versions/3.8.7/lib/python3.8/concurrent/futures/thread.py", line 57, in run
    result = self.fn(*self.args, **self.kwargs)
  File "/home/ec2-user/github/databricks-migrate/pipeline/pipeline.py", line 73, in _run_task
    task.run()
  File "/home/ec2-user/github/databricks-migrate/tasks/tasks.py", line 268, in run
    jobs_c.import_job_configs()
  File "/home/ec2-user/github/databricks-migrate/dbclient/JobsClient.py", line 197, in import_job_configs
    self.update_imported_job_names(error_logger, checkpoint_job_configs_set)
  File "/home/ec2-user/github/databricks-migrate/dbclient/JobsClient.py", line 73, in update_imported_job_names
    resp = self.post('/jobs/update', update_args)
  File "/home/ec2-user/github/databricks-migrate/dbclient/dbclient.py", line 223, in post
    return self.http_req('post', endpoint, json_params, version, print_json, files_json)
  File "/home/ec2-user/github/databricks-migrate/dbclient/dbclient.py", line 207, in http_req
    raise Exception("Error: {0} request failed with code {1}\n{2}".format(http_type,
Exception: Error: post request failed with code 429
```

I think it is important to note that we failed in this step:
```
File "/home/ec2-user/github/databricks-migrate/dbclient/JobsClient.py", line 73, in update_imported_job_names
    resp = self.post('/jobs/update', update_args)
```

Ok ... so time to give check pointing a true test.
I re-executed with same command:
```
date; time python migration_pipeline.py --import-pipeline --profile scratch-green --session M20220310200413 --use-checkpoint --debug --num-parallel 10 --skip-tasks 'metastore' 'metastore_table_acls' 'workspace_acls' 'notebooks' 2>&1 | tee -a tmp.txt
```

This second run failed with a new exception:
```
Traceback (most recent call last):
  File "migration_pipeline.py", line 328, in <module>
    main()
  File "migration_pipeline.py", line 324, in main
    pipeline.run()
  File "/home/ec2-user/github/databricks-migrate/pipeline/pipeline.py", line 64, in run
    future.result()
  File "/home/ec2-user/.pyenv/versions/3.8.7/lib/python3.8/concurrent/futures/_base.py", line 439, in result
    return self.__get_result()
  File "/home/ec2-user/.pyenv/versions/3.8.7/lib/python3.8/concurrent/futures/_base.py", line 388, in __get_result
    raise self._exception
  File "/home/ec2-user/.pyenv/versions/3.8.7/lib/python3.8/concurrent/futures/thread.py", line 57, in run
    result = self.fn(*self.args, **self.kwargs)
  File "/home/ec2-user/github/databricks-migrate/pipeline/pipeline.py", line 73, in _run_task
    task.run()
  File "/home/ec2-user/github/databricks-migrate/tasks/tasks.py", line 268, in run
    jobs_c.import_job_configs()
  File "/home/ec2-user/github/databricks-migrate/dbclient/JobsClient.py", line 187, in import_job_configs
    current_job_id = job_id_by_name[acl_conf['job_name']]
KeyError: 'dse_ds_ganges_daily_parse:::21643'
```
I think what has happened is the `update_imported_job_names` function has renamed some of the jobs and then on subsequent checkpointed run, the `get_job_id_by_name` function is returning the new job name for those jobs that had been renamed on the first execution.

kevinkim9264:
@paulg-mscience Good test!
So the first one failed.. obviously because of the rate limiting.
The second try scenario is very weird but here is what probably happened:
First try:
1. On the ACL update stage, some jobs failed due to rate limiting.
2. we do `logging_utils.log_reponse_error` but don't immediately fail. Moves onto the update imported job names stage.
3. updated all or partial of the job names.
4. Toward the end, before moving to the next **task**, the pipeline fails because error_logger has some errors.

Second try:
1. On the ACL update stage, it gets the current job list. Because on the first try, partial or all of the renames worked, some jobs are already renamed and hence do not have :::xxxx part. 
2. acl_conf['job_name'] always has `job_name:::xxxx` but the some or all of the current jobs list only have `job_name` without :::xxxx because they are already renamed on the first try. 
3. hence `job_id_by_name[acl_conf['job_name']]` fails.

On the ACL apply stage, I think it makes sense to do a bandage of checking whether the acl_conf['job_name'] exists in the map. If it doesn't, fall back to job_name (without :::xxxx) and if job_name doesn't exist, skip. (because it means the job import itself failed on the previous stage)

Let me put up the fix asap

kevinkim9264:
@paulg-mscience added the exiting logic. This is not clean. The better way is to actually split the job_import, job_acl_apply, job_rename to be on separate tasks. But this short term fix should do the trick

paulg-mscience:
I am not so sure about the exit fast code. Let me think about that for a bit.

paulg-mscience:
> On the ACL apply stage, I think it makes sense to do a bandage of checking whether the acl_conf['job_name'] exists in the map. If it doesn't, fall back to job_name (without :::xxxx) and if job_name doesn't exist, skip. (because it means the job import itself failed on the previous stage)

@kevinkim9264 I prefer the above mentioned fix versus your recent fail fast fix. Although it is a bit hacky, I believe it will work. The fail fast approach will require users to research and modify log files in order to proceed with the import. That could be error prone and time consuming.

kevinkim9264:
@paulg-mscience exit fast more ideal and is more similar to the other suggestion of splitting them into 3 separate stages.
By failing fast, basically we will never get into a state where `job import succeeded, job acl failed, and job rename succeeded` which is exactly the case that would cause the problem you saw above. 

Exit fast will not require users to modify log files because it will always be the case where job_rename will be called only if job_acl apply is successful which is called only if job_import is successful.

paulg-mscience:
> Exit fast will not require users to modify log files because it will always be the case where job_rename will be called only if job_acl apply is successful which is called only if job_import is successful.

@kevinkim9264 Ahhh ... I finally understand. I will test your fail fast logic on Monday. Thanks!

kevinkim9264:
Thank you for your help!

dbclient/ClustersClient.py
```diff
@@ -230,10 +234,16 @@ def import_cluster_configs(self, log_file='clusters.log', acl_log_file='acl_clus
...
+        error_logger = logging_utils.get_error_logger(
+            wmconstants.WM_IMPORT, wmconstants.CLUSTER_OBJECT, self.get_export_dir())
+        checkpoint_cluster_configs_set = self._checkpoint_service.get_checkpoint_key_set(
+            wmconstants.WM_IMPORT, wmconstants.CLUSTER_OBJECT)
         # get instance pool id mappings
         with open(cluster_log, 'r') as fp:
             for line in fp:
                 cluster_conf = json.loads(line)
+                if 'cluster_id' in cluster_conf and checkpoint_cluster_configs_set.contains(cluster_conf['cluster_id']):
+                    continue
```
yubing, line 246: Log something when skipping. I think it'll help users understand the behavior. The same to other places.

dbclient/ClustersClient.py
```diff
@@ -230,10 +234,16 @@ def import_cluster_configs(self, log_file='clusters.log', acl_log_file='acl_clus
             return
         current_cluster_names = set([x.get('cluster_name', None) for x in self.get_cluster_list(False)])
         old_2_new_policy_ids = self.get_new_policy_id_dict()  # dict of {old_id : new_id}
+        error_logger = logging_utils.get_error_logger(
+            wmconstants.WM_IMPORT, wmconstants.CLUSTER_OBJECT, self.get_export_dir())
+        checkpoint_cluster_configs_set = self._checkpoint_service.get_checkpoint_key_set(
```
yubing, line 239: Note that you are reusing the same checkpoint `CLUSTER_OBJECT` for different objects. If the ids happen to be the same, there will be data loss.

dbclient/JobsClient.py
yubing: @kevinkim9264 I think it'll will helpful to add your explanation about str() to code comments.

dbclient/JobsClient.py
```diff
@@ -116,6 +121,8 @@ def import_job_configs(self, log_file='jobs.log', acl_file='acl_jobs.log'):
         # get an old cluster id to new cluster id mapping object
         cluster_mapping = self.get_cluster_id_mapping()
         old_2_new_policy_ids = self.get_new_policy_id_dict()  # dict { old_policy_id : new_policy_id }
+        checkpoint_job_configs_set = self._checkpoint_service.get_checkpoint_key_set(
```
yubing, line 124: The same problem.

dbclient/ClustersClient.py
kevinkim9264: yeah, this is rather intentional as the ids of these objects cannot be same. (just different id formats to start with)
Didn't want to convolute the number of files. But please let me know if you think the otherwise, can definitely make the change.

dbclient/ClustersClient.py
kevinkim9264: Make sesnse!

dbclient/JobsClient.py
kevinkim9264: ditto

dbclient/ClustersClient.py
kevinkim9264: Actually, it's already logged as part of the "contains" message, just like other checkpointer logics

dbclient/ClustersClient.py
kevinkim9264: e.g. one is UUID, the other is cluster path.

yubing (approved):

PR #142, opened 2022-03-16
[Issue 126] Add http retry logic

paulg-mscience:
This PR supports http retry logic for making requests to the Databricks API.

The following two command line parameters were added:

```
--retry-total RETRY_TOTAL
Total number or retries when making calls to Databricks API

--retry-backoff RETRY_BACKOFF
Backoff factor to apply between retry attempts when making calls to Databricks API
```

This PR also refactored the code to propagate exceptions that are thrown by threads.
Unit test: `python -m unittest discover -s test -p 'thread*_test.py'`

Manual test of export which successfully exported 90,000 notebooks.
```
date; time python migration_pipeline.py --export-pipeline --profile old --use-checkpoint --debug --num-parallel 20 --retry-total=20 --retry-backoff=1.0 --notebook-format SOURCE --skip-tasks 'metastore' 'metastore_table_acls' 2>&1 | tee -a tmp.txt
```

Manual test of import which successfully imported 90,000 notebooks with many retries of http 429:
```
date; time python migration_pipeline.py --import-pipeline --profile scratch-green --session M20220311165759 --archive-missing --use-checkpoint --debug --num-parallel 8 --retry-total=20 --retry-backoff=1.0 --notebook-format SOURCE --overwrite-notebooks --skip-tasks 'metastore' 'metastore_table_acls' 'workspace_acls' 2>&1 | tee -a tmp.txt
```

### DIFF
--- a/dbclient/dbclient.py
+++ b/dbclient/dbclient.py
@@ -37,10 +37,17 @@ class dbclient:
     Rest API Wrapper for Databricks APIs
     """
     # set of http error codes to retry
-    http_retry_codes = [429, 503]
+    # 429: Too Many Requests
+    # 503: Service Unavailable
+    # 504: Gateway Timeout
+    http_retry_codes = [429, 503, 504]
+
     # set of http error codes to throw an exception if hit. Handles client and auth errors.
     # Also include the retry error codes in case all retry attempts fail.
-    http_error_codes = [401, 500, 502, 504] + http_retry_codes
+    # 401: Unauthorized
+    # 500: Internal Server Error
+    # 502: Bad Gateway
+    http_error_codes = [401, 500, 502] + http_retry_codes
 
     def __init__(self, configs):
         self._profile = configs['profile']
--- a/dbclient/dbclient.py
+++ b/dbclient/dbclient.py
@@ -1,11 +1,14 @@
 import json
 import os
 import requests
+from requests.adapters import HTTPAdapter
 import fileinput
 import re
 from dbclient import parser
 import time
 import requests.packages.urllib3
+from requests.packages.urllib3 import Retry
+import threading
 import logging_utils
 import logging
 
@@ -33,8 +36,11 @@ class dbclient:
     """
     Rest API Wrapper for Databricks APIs
     """
-    # set of http error codes to throw an exception if hit. Handles client and auth errors
-    http_error_codes = [401, 429]
+    # set of http error codes to retry
+    http_retry_codes = [429, 503]
+    # set of http error codes to throw an exception if hit. Handles client and auth errors.
+    # Also include the retry error codes in case all retry attempts fail.
+    http_error_codes = [401, 500, 502, 504] + http_retry_codes
 
     def __init__(self, configs):
         self._profile = configs['profile']
@@ -49,6 +55,11 @@ class dbclient:
         self._verify_ssl = configs['verify_ssl']
         self._file_format = configs['file_format']
         self._is_overwrite_notebook = configs['overwrite_notebooks']
+        self._local = threading.local()
+        self._retry_total = configs['retry_total']
+        self._retry_backoff = configs['retry_backoff']
+        if configs['debug']:
+            logging.getLogger("urllib3").setLevel(logging.DEBUG)
         if self._verify_ssl:
             # set these env variables if skip SSL verification is enabled
             os.environ['REQUESTS_CA_BUNDLE'] = ""
@@ -103,7 +114,7 @@ class dbclient:
         if self._url[-4:] != '.com' and self._url[-4:] != '.net':
             print("Hostname should end in '.com'")
             return -1
-        results = requests.get(self._url + '/api/2.0/clusters/spark-versions', headers=self._token,
+        results = self.req_session().get(self._url + '/api/2.0/clusters/spark-versions', headers=self._token,
                                verify=self._verify_ssl)
         http_status_code = results.status_code
         if http_status_code != 200:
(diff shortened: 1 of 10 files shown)


paulg-mscience:
@kevinkim9264 This PR has been rebased and re-tested and is ready for review.

paulg-mscience:
@fstraulemann, @goodwillpruning ... fyi

threading_utils.py
```diff
@@ -0,0 +1,3 @@
+def propagate_exceptions(futures):
```
kevinkim9264, line 1: Thanks for doing this!

threading_utils.py
```diff
@@ -0,0 +1,3 @@
...
+    # Calling result() on a future whose execution raised an exception will propagate the exception to the caller
+    [future.result() for future in futures]
```
kevinkim9264, line 3: nit: can we add a new line at the end of the file?

dbclient/dbclient.py
```diff
@@ -33,8 +36,11 @@ class dbclient:
     """
     Rest API Wrapper for Databricks APIs
     """
-    # set of http error codes to throw an exception if hit. Handles client and auth errors
-    http_error_codes = [401, 429]
+    # set of http error codes to retry
+    http_retry_codes = [429, 503]
```
kevinkim9264, line 40: on `http_retry_codes` and `http_error_codes`, can we leave comments on what each of the code we are handling? (for more explicit readability)

dbclient/dbclient.py
```diff
@@ -146,6 +157,24 @@ def _should_retry_with_new_token(self, raw_results):
...
+        """
+        Creates one new request session per thread with retry adapater
+        """
+        if not hasattr(self._local, "session"):
+            adapter = HTTPAdapter(max_retries=Retry(
+                total=self._retry_total, 
+                backoff_factor=self._retry_backoff,
+                status_forcelist=self.http_retry_codes,
+                allowed_methods=frozenset({'DELETE', 'GET', 'HEAD', 'OPTIONS', 'PATCH', 'POST', 'PUT', 'TRACE' }),
+                raise_on_status=False
```
kevinkim9264, line 170: Have we tested where the retry was exhausted? With this boolean set to False, it just returns a response and be logged into error logger and continue on the next calls. 

I wonder if we should just throw instead and stop the task if it fails even with all the number of retries
WDYT?

dbclient/dbclient.py
kevinkim9264: hmm, if we are retrying on 503, shouldn't we also retry on 504?

kevinkim9264 (changes requested):
Thanks for the PR! Looks good in general
@mrchristine could you take a look as well?

threading_utils.py
paulg-mscience: fixed https://github.com/databrickslabs/migrate/pull/142/commits/dcad022083961b4826befbf23376145b1641896f

dbclient/dbclient.py
paulg-mscience: fixed https://github.com/databrickslabs/migrate/pull/142/commits/6f05b385f2f470223f61c712d5651cdb59cea5dd

dbclient/dbclient.py
paulg-mscience: I think we are good here.

I have tested the case where retries are exhausted. In this case it returns the response with the offending http status code and the calling code behaves as if retry were not being used. I like this behavior as it is not disruptive to the current logic.

When I tested with exhausting the retries while importing notebooks, the http status code was propagated and re-thrown by the calling code because 429 is in the list of `http_error_codes`.  From here I could adjust my `--retry-total` command line arg and re-run picking back up at my last notebook checkpoint.

paulg-mscience:
@kevinkim9264 Thank you for the code review. Let me re-test my changes. I will ping when finished testing.

dbclient/dbclient.py
kevinkim9264: sounds good to me

paulg-mscience:
@kevinkim9264 Testing went well. Ready for another code review.

test/threading_utils_test.py
```diff
@@ -0,0 +1,25 @@
...
+
+        def run_stuff():
+            fut1 = ThreadPoolExecutor(2).submit(do_something_good)
+            fut2 = ThreadPoolExecutor(2).submit(do_something_bad)
+            return fut1, fut2
+
+        with self.assertRaises(MyBadException):
+            futures = run_stuff()
+            concurrent.futures.wait(futures)
+            propagate_exceptions(futures)
```
kevinkim9264, line 25: nit: newline at the end of file.

threading_utils.py
```diff
@@ -0,0 +1,4 @@
...
+
```
kevinkim9264, line 4: would be great if we could get rid of the github no newline indication
(if you have no character at line 4, I think it will not have it)

kevinkim9264 (approved):
lgtm

threading_utils.py
paulg-mscience: fixed ... https://github.com/databrickslabs/migrate/pull/142/commits/0c2314d44698baad41fad13298cef7e8c53834fe

paulg-mscience:
@mrchristine, @kevinkim9264 Will you be merging this PR this week? :)

kevinkim9264:
@paulg-mscience yup, def will do! Waiting for @mrchristine or @gregwood-db to give a review.

gregwood-db:
@kevinkim9264 @paulg-mscience just took a look through the updates- all looks good to me. Thanks for the work on this!

kevinkim9264:
Merged. @paulg-mscience

PR #145, opened 2022-03-21
[PLAT-46723] ML run import  (with run artifact download and upload)

kevinkim9264:
This PR supports ML run import **with artifacts**. 

Workflow:
`python3 import_db.py --profile mwc_dst --src-profile mwc_src --mlflow-runs --use-checkpoint --num-parallel 4` command (note that it requires **--src-profile** flag) does the following:
1. Get all the exported runs from the `mlflow_runs.db` file. (Fetch 10000 runs per time for the sake of memory)
2. For the fetched runs, parallelize and call `self._create_run_and_log` for each of the runs.
==== Parallelize ========
4. `_create_run_and_log` a) creates a run b) log the metrics, params, and tags c) download and upload the run artifact (src_profile is needed for this step)  between a,b, and c steps, we checkpoint the individual operations as well via `steps_checkpointer` (which is different from the master checkpointer)
5. if the run has "mlflow.parentRunId", before doing the current run's operation, recursively call `_create_run_and_log` on the parent run, which should return the newly created parent run Id, which can be used for the current run's parentId tag.
6. Checkpoint the successfully imported run (master checkpointer). The checkpoint will have old_run_id -> new_run_id mapping.
===== Parallelize done =====
8. Copy the checkpointer file to `mlflow_runs_id_map.log` 

The PR also does the followings:
1. Introduce CheckpointKeyMap class, which is similar to CheckpointKeySet, but instead of saving a key, it saves key, value pairs. This is useful when we need to return the value upon running into checkpointed key. (e.g. In the case of ml run import, _create_experiment must return the new id. This case is necessary if the parent run's import was successful but the children's import was transiently failed/interrupted which requires rerun. Upon the rerun, the checkpoint should return the new id of the already imported parent run, so that the children can use that new id as their parent id as they are getting imported.

2. Make import_experiments more robust by not failing if the same experiment already exists in the destination workspace, in which case the script gets the experiment Id and logs to the experiment_id_map log file.

Manual test:
```
python3 export_db.py --profile mwc_src --mlflow-runs --use-checkpoint --num-parallel 4
python3 import_db.py --profile mwc_dst --src-profile mwc_src  --mlflow-runs --use-checkpoint --num-parallel 4
```
Src:
Screen Shot 2022-03-21 at 4 23 06 PM

Dst:
Screen Shot 2022-03-21 at 4 23 16 PM
(confirmed that the parents are properly set)

Importing ml runs fails if the experiment_id the run belongs to doesn't exist in the experiment_id_map file. (it means, some of the experiment import must've failed):
```
# some experiment import fails
2022-03-21,10:47:19;ERROR;{'error_code': 'RESOURCE_DOES_NOT_EXIST', 'message': 'Parent directory /Repos/yonghong.goh+tzar@databricks.com/databricks_repo/demos does not exist.'}


# Some runs import fail, because the experiment_id doesn't exist.
2022-03-21,11:22:55;ERROR;Run: 7228502ea0a84d4e8269646e988b3b24 originally belongs to experiment_id 904276049961994, but 904276049961994 does not exist in mlflow_experiments_id_map.log. Make sure the experiment
is correctly imported before importing runs.
2022-03-21,11:22:55;ERROR;Run: fa9d39e2041e4e33a6bf9703df85c9a0 originally belongs to experiment_id 904276049961994, but 904276049961994 does not exist in mlflow_experiments_id_map.log. Make sure the experiment
is correctly imported before importing runs.

```

Unit tests:
```
python3 -m unittest dbclient/test/MLFlowClientTest.py
python3 -m unittest test/checkpoint_service_test.py
```

### DIFF
--- a/checkpoint_service.py
+++ b/checkpoint_service.py
@@ -1,6 +1,7 @@
 import os
 from abc import ABC, abstractmethod
 import logging
+import json
 from thread_safe_writer import ThreadSafeWriter
 
 class AbstractCheckpointKeySet(ABC):
@@ -9,6 +10,18 @@ class AbstractCheckpointKeySet(ABC):
     @abstractmethod
     def write(self, key):
         """Writes key into checkpoint file."""
+        pass
+
+    @abstractmethod
+    def contains(self, key):
+        """Checks if key exists in checkpoint"""
+        pass
+
+class AbstractCheckpointKeyMap(ABC):
+    """Abstract base class for checkpoint read and write."""
+    @abstractmethod
+    def write(self, key):
+        """Writes key and value into checkpoint file."""
         pass
 
     @abstractmethod
@@ -56,6 +69,44 @@ class CheckpointKeySet(AbstractCheckpointKeySet):
     def __del__(self):
         self._checkpoint_file_append_fp.close()
 
+
+class CheckpointKeyMap(AbstractCheckpointKeyMap):
+    """Deals with checkpoint read and write. Unlike CheckpointKeySet, it also saves the value.
+    Useful when the corresponding value is needed later.
+    """
+    def __init__(self, checkpoint_file):
+        """
+        :param checkpoint_file: file to read / write object keys for checkpointing
+        """
+        self._checkpoint_file = checkpoint_file
+        self._checkpoint_key_map = {}
+        self._checkpoint_file_append_fp = ThreadSafeWriter(checkpoint_file, 'a')
+        self._restore_from_checkpoint_file()
+
+    def write(self, key, value):
+        if key not in self._checkpoint_key_map:
+            self._checkpoint_file_append_fp.write(json.dumps({"key": str(key), "value": str(value)}) + "\n")
+
+    def contains(self, key):
+        exists = key in self._checkpoint_key_map
+        if exists:
+            logging.info(f"{key} found in checkpoint")
+        return exists
+
+    def get(self, key):
+        return self._checkpoint_key_map[key]
+
+    def _restore_from_checkpoint_file(self):
+        if os.path.exists(self._checkpoint_file):
+            with open(self._checkpoint_file, 'r') as read_fp:
+                for single_key_value_map_str in read_fp:
+                    single_key_value_map = json.loads(single_key_value_map_str)
+                    self._checkpoint_key_map[single_key_value_map["key"]] = single_key_value_map["value"]
+
+    def __del__(self):
+        self._checkpoint_file_append_fp.close()
+
+
 class DisabledCheckpointKeySet(AbstractCheckpointKeySet):
     """Class used to denote disabled checkpointing."""
 
@@ -64,6 +115,17 @@ class DisabledCheckpointKeySet(AbstractCheckpointKeySet):
 
     def contains(self, key):
         return False
+
+
+class DisabledCheckpointKeyMap(AbstractCheckpointKeyMap):
+    def write(self, key, value):
+        pass
+
+    def contains(self, key):
+        return False
+
+    def get(self, key):
+        raise KeyError("Checkpoint is disabled")
 
 
 class CheckpointService():
@@ -88,3 +150,9 @@ class CheckpointService():
         else:
             return DisabledCheckpointKeySet()
 
+    def get_checkpoint_key_map(self, action_type, object_type):
+        if self._checkpoint_enabled:
+            checkpoint_file = f"{self._checkpoint_dir}/{action_type}_{object_type}.log"
+            return CheckpointKeyMap(checkpoint_file)
+        else:
+            return DisabledCheckpointKeyMap()
--- a/checkpoint_service.py
+++ b/checkpoint_service.py
@@ -98,6 +98,9 @@ class CheckpointKeyMap(AbstractCheckpointKeyMap):
            if the value is not "IN_USE_BY_XXX" return True (self.contains(key))
 
         This method is thread-safe since map.setdefault(key, value) is thread-safe
+        The thread that calls this method and gets False (meaning, the value wasn't there and thus this thread is using
+        this key) must set the value of this key subsequently to make sure other threads do not wait for this key
+        forever.
         """
         in_use_str = f"IN_USE_BY_{threading.get_ident()}"
         # setdefault is thread safe, so only one thread can successfully set the value for the key.
@@ -154,7 +157,7 @@ class DisabledCheckpointKeyMap(AbstractCheckpointKeyMap):
     def contains(self, key):
         return False
 
-    def check_contains_or_mark_in_use(self, key):
+    def check_contains_otherwise_mark_in_use(self, key):
         raise NotImplementedError("Checkpoint is disabled")
 
     def get(self, key):
--- a/checkpoint_service.py
+++ b/checkpoint_service.py
@@ -105,9 +105,9 @@ class CheckpointKeyMap(AbstractCheckpointKeyMap):
         if result == in_use_str:
             return False
         else:
-            while "IN_USE_BY" in self._checkpoint_key_map[key]:
+            while key in self._checkpoint_key_map and "IN_USE_BY" in self._checkpoint_key_map[key]:
                 logging.info(f"Waiting for {key} result to be available..")
-                time.sleep(2)
+                time.sleep(5)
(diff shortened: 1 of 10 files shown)


dbclient/MLFlowClient.py
```diff
@@ -179,8 +155,21 @@ def _create_experiment(self, experiment_str, id_map_writer, checkpointer, error_
...
             logging.info(f"Successfully created experiment with name: {name}. id: {new_id}")
         except RestException as error:
-            logging.info(f"create experiment failed for id: {id}, name: {name}. Logging it to error file..")
-            error_logger.error(error.json)
+            # If the resource already exists, then we can consider it successful and checkpoint it.
+            if error.json['error_code'] == 'RESOURCE_ALREADY_EXISTS':
+                logging.info(error.json['message'] + f" Trying to get the experiment_id of the existing experiment: {name}...")
+                try:
+                    new_id = self.client.get_experiment_by_name(name).experiment_id
+                    logging.info(f"Successfully retrieved an id: {new_id} for experiment: {name} in the target workspace.")
```
smurching, line 163: We should also just make sure to set tags in this case, e.g. for notebook experiments

dbclient/MLFlowClient.py
```diff
@@ -179,8 +155,21 @@ def _create_experiment(self, experiment_str, id_map_writer, checkpointer, error_
             new_id = self.client.create_experiment(name, artifact_location, dict_tags)
             logging.info(f"Successfully created experiment with name: {name}. id: {new_id}")
         except RestException as error:
-            logging.info(f"create experiment failed for id: {id}, name: {name}. Logging it to error file..")
-            error_logger.error(error.json)
+            # If the resource already exists, then we can consider it successful and checkpoint it.
+            if error.json['error_code'] == 'RESOURCE_ALREADY_EXISTS':
```
smurching, line 159: Should we also check here whether the experiment is a notebook experiment?

kevinkim9264: We shouldn't. If even if is non-notebook experiment, we can't create the experiment anyway. If the experiment already exists in the same path, it's what it is.

dbclient/MLFlowClient.py
kevinkim9264: done!

checkpoint_service.py
```diff
@@ -56,6 +71,72 @@ def _restore_from_checkpoint_file(self):
...
+        self._restore_from_checkpoint_file()
+
+    def write(self, key, value):
+        if key not in self._checkpoint_key_map or "IN_USE_BY" in self._checkpoint_key_map[key]:
+            self._checkpoint_key_map[key] = value
+            self._checkpoint_file_append_fp.write(json.dumps({"key": str(key), "value": str(value)}) + "\n")
+
+    def check_contains_otherwise_mark_in_use(self, key):
+        """
+        If the key_map does not have the key value yet, mark the key to be IN_USE_BY_$THREAD_ID, and return False.
```
smurching, line 95: Why do we mark as `IN_USE_BY_$THREAD_ID`? Is the expectation that the current thread will subsequently populate & set the value?

smurching: That seems reasonable but might just be good to have a comment to clarify

dbclient/MLFlowClient.py
```diff
@@ -198,3 +192,176 @@ def _cleanse_artifact_location(self, artifact_location):
...
+        con = sqlite3.connect(mlflow_runs_file)
+        cur = con.execute("SELECT * FROM runs")
+        # TODO(kevin): make this configurable later
+        runs = cur.fetchmany(10000)
+        while(len(runs) > 0):
+            with ThreadPoolExecutor(max_workers=num_parallel) as executor:
+                # run_id = run[0]
+                # start_time = run[1]
+                # run_obj = json.loads(run[2])
+                futures = [executor.submit(self._create_run_and_log, src_client, mlflow_runs_file, run[0], run[1], json.loads(run[2]), experiment_id_map, self.export_dir + ml_run_artifacts_dir, error_logger, mlflow_runs_checkpointer, mlflow_runs_steps_checkpointer) for run in runs]
```
smurching, line 239: BTW out of curiosity why is the run-based locking necessary? I would expect that when using `ThreadPoolExecutor` no two threads would ever attempt to migrate the same run

smurching: It seems we could handle that by grouping together trees of runs by parent ID first before attempting to import them (so that `r1` is guaranteed to be imported before `r2`). We could also then have an individual thread handle importing a tree of runs without worrying about race conditions potentially. But the current approach seems ok too since it does guarantee that parents are always created before children

dbclient/MLFlowClient.py
```diff
@@ -198,3 +192,176 @@ def _cleanse_artifact_location(self, artifact_location):
...
+        # workspace. After uploading, remove the temp dir along with the artifacts.
+        temp_artifact_dir = ml_run_artifacts_dir + old_run_id + "_temp/"
+        shutil.rmtree(temp_artifact_dir, ignore_errors=True)
+        os.makedirs(temp_artifact_dir)
+        artifacts = src_client.list_artifacts(old_run_id)
+        if len(artifacts) == 0:
+            return
+
+        logging.info(f"Downloading run artifacts for run_id: {old_run_id}")
+        src_client.download_artifacts(old_run_id, "", temp_artifact_dir)
```
smurching, line 352: nit: may want to use a `try / finally` block here to make sure we always call `shutil.rmtree(temp_artifact_dir)` even if the upload attempt fails in the current thread

smurching (approved):
Left few questions/minor comments, LGTM to merge once those are addressed!

checkpoint_service.py
kevinkim9264: yes exactly. This is needed because of the behavior we desire which is to make one thread be responsible for a runs and all of its parent runs. In this behavior we can run into this situation where:
1. Thread A is assigned  run A whose parents are run E -> run T
2. Thread B is assigned run T

Case 1 is when thread A takes run T before Thread B gets there.
Thread A takes and mark the run T's value to be `IN_USE_BY_$THREAD_ID_OF_THREAD_A`
Thread B tries to take but sees that it is in Use. It has to wait until the real value is available because it has to return the run Id.
Thread A finishes the work, and correctly sets the value to be the new run Id.
Thread B sees the new value and return. 

Case 2 is when thread B takes run T before Thread A gets there.
Thread B takes and mark the run T's value to be `IN_USE_BY_$THREAD_ID_OF_THREAD_B`
Thread A tries to take but sees that it is in Use. It has to wait until the real value is available.
Thread B finishes the work and sets the value to be the the new run Id.
Thread A sees the new value and return **which is used as run E's parent run Id as the recursion returns.**

dbclient/MLFlowClient.py
kevinkim9264: Yes your approach would work too, but the problem is that `grouping together trees of runs by parent ID first` is not that easy in terms of the performance. In order to do that, we would have to go through all the runs first and **load it in to memory**. Note that we are already chunking the fetchmany by only certain number, but there is no guarantee that all the same rooted trees are in the same chunk. This requires us to load all the data in memory to do the grouping.

Another similar approach is to recognize this at the export time and indicate it in database schema. (by setting a group or something). We can set the list of group Ids in a separate file, and on import, we can probably parallelize across those different group Ids. (where each of the thread does SELECT * FROM RUNS WHERE GROUP_ID=$GROUP_ID).

The current method is complicated but good in a sense that program can do blind parallelization and just wait if it ever runs into a situation where its parernt run is being handled by other threads. The key is to ensure the atomicity of `check_contains_otherwise_mark_in_use` function which is true by testing and the document explanation.

checkpoint_service.py
kevinkim9264: let me leave the comment that the whoever is calling this to set, should also set the value later

dbclient/MLFlowClient.py
kevinkim9264: good point. Done!

kevinkim9264:
@gregwood-db I will merge the ML PRs without your stamp for now. I will have to create another PR that merges tiger into master, so you can look at the codes there. We can set up a quick meeting to go over the codes together.

PR #147, opened 2022-03-23
[Issue 139] Set job owner to admin user when original job owner is obsolete user

paulg-mscience:
This PR handles setting ACLs for a job whose original owner was an obsolete user. In this case the job will be owned by the admin user whose credentials are tied to the databricks token that is being used to run the import.

There is also an optimization to only ask for the admin username once, instead of each time we apply job ACLs.

This PR has been manually tested by running a job export and job import.

### DIFF
--- a/dbclient/dbclient.py
+++ b/dbclient/dbclient.py
@@ -287,7 +287,7 @@ class dbclient:
         user_name = self.get('/preview/scim/v2/Me').get('userName')
         return user_name
 
-    def build_acl_args(self, full_acl_list, is_jobs=False):
+    def build_acl_args(self, full_acl_list, is_jobs=False, admin_user_name=None):
         """
         Take the ACL json and return a json that corresponds to the proper input with permission level one level higher
         { 'acl': [ { (user_name, group_name): {'permission_level': '*'}, ... ] }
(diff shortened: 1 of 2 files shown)


paulg-mscience:
Is there any desire to merge this PR? If not, I will close and keep this code in our fork.

paulg-mscience:
@kevinkim9264 Do you think this PR is generally useful? Or should I close it and keep this logic specific to our private fork?

dbclient/dbclient.py
```diff
@@ -312,9 +312,11 @@ def build_acl_args(self, full_acl_list, is_jobs=False):
                         current_owner = member.get('group_name')
 
         if is_jobs:
-            me = self.whoami()
-            if current_owner != me:
-                update_admin = {'user_name': self.whoami(),
+            # set current_owner to admin if current_owner is empty
+            # current_owner is empty when original job owner is obsolete user
+            current_owner = current_owner or admin_user_name
+            if current_owner != admin_user_name:
```
kevinkim9264, line 318: @paulg-mscience 

I am confused by this change. The expected behavior is that we should append the `update_admin` block if this script runner (whoami) is not the current owner of the job. (whether it being empty or not)

In this changed behavior, if the current_owner is empty, then we are setting `current_owner = whoami` and it will make `current_owner == admin_user_name` in which case, it won't append the `update_admin` block which is not an expected behavior.

Could you explain what's going on?

dbclient/dbclient.py
```diff
@@ -312,9 +312,11 @@ def build_acl_args(self, full_acl_list, is_jobs=False):
                         current_owner = member.get('group_name')
 
         if is_jobs:
-            me = self.whoami()
-            if current_owner != me:
-                update_admin = {'user_name': self.whoami(),
```
kevinkim9264, line 317: Wow, we should at least make it 
```
{'user_name': me} 
```

kevinkim9264 (changes requested):

PR #165, opened 2022-04-19
Import_workspace_acl should not be called in parallel.

kevinkim9264:
Databricks ACL teams have reported that parallel import for workspace ACL can cause issues to the backed services. For now, we need to disable parallel feature for workspace ACL import until further scale out / rate limit is implemented in the Databricks backend service.


iamsgarg-db (approved):
ty @kevinkim9264 for making this change.

PR #195, opened 2022-07-28
Added logging of job id mapping

sergiu-databricks:
Currently, the jobs ids do not get preserved when migrating. This change adds a log containing a mapping between job ids in the old workspace to job ids in the new workspace.


gregwood-db (approved):
LGTM- thanks for the PR @sergiu-databricks

sergiu-databricks:
@gregwood-db, I think I am going to actually cancel this PR. I realised the format of the job log is not in line with what is done for mlflow experiments. I will make a new PR with the right format for the file, as well as added functionality to solve this issue:  [Job pause_status is not migrated](https://github.com/databrickslabs/migrate/issues/196)

Issue #196, opened 2022-08-03
Job pause_status is not migrated


sergiu-databricks:
Currently there is no functionality to migrate the `pause_status` of a job. During a regular migration, all jobs are migrated with their schedule, but are set up as PAUSED. This is good, as you do not want to have the same job running in two different environments at the same time.

Afterwards, there is functionality to bulk pause/unpause all jobs, but this does what it says on the tin: apply the change to all jobs, regardless of the initial status. So if there is a mixture of paused and unpaused jobs in the original workspace, there is no current way of migrating that. This is bad.

It is worth noting that the original pause_status is stored in `jobs.log`.

sergiu-databricks:
One likely cause of this is that we do not save the mapping between job ids. This [PR](https://github.com/databrickslabs/migrate/pull/195) should solve this though, allowing for this functionality to be implemented.

gregwood-db:
Merged in #197

PR #197, opened 2022-08-03
Job pausing

sergiu-databricks:
This pull request adds two functionalities:

- adds a log containing a mapping between job ids in the old workspace to job ids in the new workspace. This is in the same format as the log id mapping used for mlflow experiments
- adds functionality to unpause only jobs with schedules which are not paused in the source workspace - see issue [196](https://github.com/databrickslabs/migrate/issues/196)


sergiu-databricks:
@gregwood-db  this is the new PR, with the right format and the added functionality

gregwood-db (approved):
Approving for merge

gregwood-db:
Approving and merging this PR- thanks @sergiu-databricks!

PR #202, opened 2022-09-12
Options to set jobs as single user or shared to allow users to change their jobs config to UC friendly configs

alex-lopes-databricks:
The motivation for these options is to allow users migrating from hive to set their jobs as single user or shared to allow them to access unity catalog tables


gregwood-db:
@alex-lopes-databricks will give this a look- mostly just need to confirm that it works with MTJ, since we've updated the tool to support those now.

PR #210, opened 2022-10-17
Using supplied AWS_PROFILE instead of hardcoded value

adiospeds:
I was receiving below error(stuck in for loop) when I tried to create a jumpbox using the script:

The config profile (aws-field-eng_databricks-power-user) could not be found Instance i-xxxxxxxxxxx is in Not Ready state. Waiting for the instance to be in running state. Checking again in 5 seconds..


gregwood-db (approved):
Thanks for catching this!

Issue #217, opened 2022-12-27
multi-task jobs fail to import


lorenzorubi-db:
when a multi-task job has several tasks that point to different `existing_cluster_id`, the tool fails to import the job with a weird error message:

```
2022-12-27,12:30:51;INFO;Current Job Name: Job name X
2022-12-27,12:30:51;WARNING;{"error_code": "INVALID_PARAMETER_VALUE", "message": "Cluster 1122-151539-wo9bvqa5 does not exist"}

2022-12-27,12:30:51;INFO;Resetting job to use default cluster configs due to expired configurations.
2022-12-27,12:30:51;WARNING;{"error_code": "INVALID_PARAMETER_VALUE", "message": "Multi-task jobs require clusterspec to be specified within the tasks."}

2022-12-27,12:30:51;ERROR;{"error_code": "INVALID_PARAMETER_VALUE", "message": "Multi-task jobs require clusterspec to be specified within the tasks.", "http_status_code": 400}
```

(it seems to happen actually due to an error when mapping the clusters, the cluster `1122-151539-wo9bvqa5` exists in the source workspace but does not exist in the target workspace)

lorenzorubi-db:
solved in [this commit](https://github.com/databrickslabs/migrate/commit/e731057044be6d84166eb83772c4a4c2266fecfa)

PR #218, opened 2022-12-27
multi-task jobs may have existing_cluster_id per task

lorenzorubi-db:
solves issue #217

### DIFF
--- a/dbclient/JobsClient.py
+++ b/dbclient/JobsClient.py
@@ -159,6 +159,7 @@ class JobsClient(ClustersClient):
                 else:
                     new_cluster_conf = cluster_conf
                 settings['new_cluster'] = new_cluster_conf
+            return settings
 
         with open(jobs_log, 'r') as fp, open(job_map_log, 'w') as jm_fp:
             for line in fp:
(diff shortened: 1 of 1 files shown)


dbclient/JobsClient.py
```diff
@@ -178,8 +179,19 @@ def adjust_ids_for_cluster(settings): #job_settings or task_settings
                 if 'format' not in job_settings or job_settings.get('format') == 'SINGLE_TASK':
                     adjust_ids_for_cluster(job_settings)
                 else:
+                    mod_task_settings = []
                     for task_settings in job_settings.get('job_clusters', []):
                         adjust_ids_for_cluster(task_settings)
+                    if mod_task_settings:
```
gregwood-db, line 185: Will this ever evaluate to true, since it's being set to [] in line 182?

lorenzorubi-db: see [here](https://stackoverflow.com/a/53522/3519000)
an empty list returns a False in python
but yes we could use `len(mod_task_settings) > 0` to make it more clear

lorenzorubi-db: added a comment that hopefully is more clear

gregwood-db (approved):

PR #222, opened 2022-12-28
Issue216 Add SPN import logic to ScimClient.py

hwang-db:
Adding import logic for Azure Service Principals. The exported groups log will contain service principal info, this added logic will detect if each member inside each group is a service principal or not, if yes, it will form a basic json object and post it via the ScimClient to import those SPNs into the target workspace. Tested with Azure Databricks.


hwang-db:
Found issue - this will create duplicated workspace directory for each SPN. Close to revise.

PR #231, opened 2023-01-10
Fix cluster OriginalCreator tag value and check 'new_cluster' config exists in task

neitomic:
This PR tent to fix 2 issues

### Cluster tag value not allow special character

The custom tag when create cluster doesn't allow special character, but the script is using email address for the `OriginalCreator` tag, which contains `@` and cause error

The PR replace the `@` with `_at_`

### `new_cluster` not exists in task settings
The script default get the new_cluster config if there is no `existing_cluster_id` in settings, but the `new_cluster` key is optional.

The PR check the `new_cluster` to make sure it exists before using it.


neitomic:
Oh duplicate with https://github.com/databrickslabs/migrate/pull/228 on the `new_cluster` part

PR #234, opened 2023-01-12
feat: add service principal option to ACLs

XaviPolaNextail:
This feat adds support for generating ACLs args for service principals, in aggregation to the already supported user & group support.

**Current behaviour**

Given a ACL Json with permissions around a service_principal,

When attempts to attach the object name and permissions,

Then an KeyError raises at attempting to search for the key "group_name" in the `else` block

**Expected behaviour**

Given a ACL Json with permissions around a service_principal,

When attempts to attach the object name and permissions,

Then it maps properly the service principal name and its permissions

### DIFF
--- a/dbclient/dbclient.py
+++ b/dbclient/dbclient.py
@@ -334,6 +334,9 @@ class dbclient:
                                   'permission_level': permissions})
                 if permissions == 'IS_OWNER':
                     current_owner = member.get('user_name')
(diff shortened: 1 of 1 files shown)


gregwood-db (approved):

gregwood-db:
Just closing as this was handled via #239- thanks for the PR!

PR #236, opened 2023-01-16
Export and import workspace folders

ferminho:
Previously, only workspace folders with notebooks are migrated, discarding:

1. empty directories
2. directories with just experiments
3. directories containing just directories like 1. and 2.

Drawbacks of that situation:

- Users expect to have a 1:1 copy of their original workspace content
- Errors in failed_import_acl_directories.log

With this PR, we ensure all workspace folders are migrated, via

- A new export task that creates the folder structure in artifacts/
- A conceptual generalization of the import notebooks functionality to import "workspace". It was already prepared to create the full directory structure, simply missing those directories in artifacts/


ferminho:
Sorry, will re-open after cleaning a couple commits that shouldn't be there

PR #237, opened 2023-01-16
Map service principals to existing ones by name

ferminho:
If the target workspace has terraformed resource such as service principals, the migration tool should try to use those instead of creating new ones. This situation doesn't happen with users, since when they're created with an already existing email Databricks API doesn't create new, extra users.

This PR introduces an opt-in flag in the migration pipeline entry point that will search for existing service principals by displayName and map them.
Enabling it will make the app double check there are no displayName duplicates in origin or destination


ferminho:
misclick

PR #240, opened 2023-01-21
Map service principal cluster acls

ferminho:
Using previous functionality, map cluster ACLs containing service principals


ferminho:
Sorry, misclick again :cry:

PR #268, opened 2023-05-23
updating validation pipeline to respect skip-tasks and keep-tasks arguments

boydZD:
- The current validation_pipeline implementation does not respect the skip-tasks and keep-tasks arguments. This causes errors in the validation_pipeline when tasks have been skipped while exporting from the source and target workspaces for comparison. 
- This PR intends to fix that following the same patterns applied for import/export keep and skip task arguments.


boydZD:
@databrickslabs/codereview please review

